PR #2, opened 2019-12-22
increase coverage on core.py and xml.py with specially crafted HTML files

vbarbaresi:
Depends on #1, to be merged after

I tried to reach some corner cases that were not covered
Some portions of the code still seem unreachable though, like 
core.py:436:
```python
            elif element.tag == 'hi':
                if element.text is not None or element.tail is not None:
```
All the <hi> tags are cleared on some step before

- Added a test for `txttocsv()` and changed it to use string formatting instead of string concat, which is not efficient (it creates a new string on each `+` operation)


vbarbaresi:
Wow it's so weird, I had bugs on GitHub on your repository, couldn't edit my pull requests and view the complete diffs

It turns out it was uBlock origin that was blocking some requests containing "adbar", because it's ony easylist https://github.com/easylist/easylist/blob/8a1791bc8b7918bbc9f14f26611443e3d2477ea9/easylist/easylist_general_block.txt#L1651

PR #13, opened 2020-05-19
Fix: ensuring processed_element is declared

jaesivsm:
If the first condition of the `handle_formatting` func is not passing, processed_element is actually undefined.


adbar:
Thanks!

PR #35, opened 2020-11-22
Add extract images feature

phongtnit:
Hello,

I added extracted images in an article with include_images option in extract() and bare_extraction() function. Please consider to accept this feature. I want to contribute and work with your package.

This feature only support in XML output format. Sorry for my other PR don't work with tests/realworld_tests.py, I fixed it.

Many thanks for your awesome package,


trafilatura/xml.py
```diff
@@ -49,7 +49,7 @@ def build_json_output(docmeta):
 def clean_attributes(tree):
     '''Remove unnecessary attributes'''
     for elem in tree.iter():
-        if elem.tag not in ('del', 'hi'):
+        if elem.tag != 'image' and elem.tag not in ('del', 'hi'):
```
adbar, line 52: should be `elem.tag not in ('del', 'hi', 'image')`

trafilatura/utils.py
```diff
@@ -259,3 +259,23 @@ def trim(string):
...
     except TypeError:
         return None
+
+
+def is_image_file(imageSrc):
+    # Regex to check valid image file extension.
+    regex = "([^\\s]+(\\.(?i:)(jpe?g|png|gif|bmp)))"
+
+    # Compile the ReGex
+    p = re.compile(regex)
```
adbar, line 269: regex and regex compilation could be faster if initialized out of the function

adbar:
Thanks for the PR, please review the two lines I commented. It would be great if it you'd write some tests as well, I can also handle it if my test scripts are too messy.

phongtnit:
> Thanks for the PR, please review the two lines I commented. It would be great if it you'd write some tests as well, I can also handle it if my test scripts are too messy.

Thanks for your feedback. I think it will be better if you write a test with basic testcases for this feature because of keeping same structure and format of all Trafilatura tests. After that, I can add some special cases of extracting images in your test.

adbar:
I believe the PR is ready to be merged. Please note that a regex targeting a few file types only may not be the best way to check for image extensions, a proper cleaning and `string.endswith(...)` could be an option, as well as a full range of extensions.

PR #44, opened 2020-12-16
Fixing KeyError raised when link[rel=canonical] has no href

Yomguithereal:
Hello @adbar,

I am opening a PR to fix the fact that a `KeyError` is throw when a page has a canonical link tag but no href. I also added a realworld case to the unit tests to take this into account. This is a webpage rendered client side by angular whose canonical link href is rendered by JavaScript (which is a bad idea on so many levels but you know the web...). Thus the test is a bit off since result is `None` due to the fact that the page does not have content before client-side rendering. I can transform this into a test checking it does not raise or find or even craft an example returning content but displaying the same lack of href. Tell me.

Have a good day,


adbar:
Good catch, thanks!

PR #50, opened 2021-01-03
Fix "from trafilatura import bare_extraction"

andreage:
from trafilatura import bare_extraction -> from trafilatura.core import bare_extraction


adbar:
Hi, thanks for the input!
There was a short-lived mismatch between docs and latest version which is now resolved with the present release (0.7.0), so I'll have to drop the pull request.

andreage:
Perfect, thank you!

PR #83, opened 2021-06-07
docs fix tutorial2 python example

### LINKED ISSUE #82: Broken docs python example
["Producing TEI files"](https://trafilatura.readthedocs.io/en/latest/tutorial2.html#producing-tei-files) docs have this example

```python3
# load the necessary components
import trafilatura
# open a file and parse it
downloaded = trafilatura.fetch_url('https://github.blog/2019-03-29-leader-spotlight-erin-spiceland/')
result = trafilatura.extract(downloaded, tei_output=True, tei_validation=True)
```

[`extract` ](https://github.com/adbar/trafilatura/blob/d26d2136307261aa3537bc8da2a26fe2c6975ae9/trafilatura/core.py#L761)function doesn't have `tei_output` flag.

To get the result as `TEI`, [`output_format`](https://github.com/tomwojcik/trafilatura/blob/master/trafilatura/core.py#L600
) has to be passed accordingly.

tomwojcik:
fixes #82

### DIFF
--- a/docs/tutorial2.rst
+++ b/docs/tutorial2.rst
@@ -16,7 +16,7 @@ In Python:
     import trafilatura
     # open a file and parse it
     downloaded = trafilatura.fetch_url('https://github.blog/2019-03-29-leader-spotlight-erin-spiceland/')
-    result = trafilatura.extract(downloaded, tei_output=True, tei_validation=True)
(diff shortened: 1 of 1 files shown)


adbar:
Thanks!

PR #91, opened 2021-07-12
Search tag by ID author, Remove roles, publications from author name.

felipehertzer:
First thank you very much for developing this package. I have an application that depends a lot on the authors' names, and this package solves my problems. However I found some difficulties to find the correct name on some sites.

[Alaska Highway News](https://www.alaskahighwaynews.ca/the-mix/the-evolving-etiquette-of-asking-about-someones-covid-19-vaccine-status-3946790) is using HTML id to markup author name `<span id="author">The Canadian Press</span>`
Before the fix: `"author": null`
After the fix: `"author": "The Canadian Press"`

[DailyMail](https://www.dailymail.co.uk/news/article-9777531/Coronavirus-Australia-Project-hosts-SLAM-government-Covid-vaccine-ad-woman-ventilator.html) has more information than the name `Levi Parsons For Daily Mail Australia`
Before the fix: `"author": "Levi Parsons For Daily Mail Australia"`
After the fix: `"author": "Levi Parsons"`

In this fix I added:
- Support for HTML id search
- I modified the regex to remove words after 'on', 'am' and 'for'


trafilatura/core.py
```diff
@@ -48,7 +48,7 @@ def handle_titles(element, dedupbool, config):
     # children
     else:
         title = deepcopy(element)
-        for child in element.iter():
+        for child in element.getchildren():
```
adbar, line 51: Hi @felipehertzer, I don't see why this change is necessary here, it doesn't affect the extraction of authors.

adbar:
Hi @felipehertzer, thank you for your effort, I only have one question about your PR (see above).
The extraction regexes are now optimized for English and German, not necessarily for other languages, but I'm fine with that for now.

trafilatura/core.py
felipehertzer: Sorry, I forgot to comment the reason for this change. Some websites like [Moree](https://www.moreechampion.com.au/story/7328821/scamwatch-book-for-your-vaccine-via-a-genuine-website/?src=rss) never finish extracting the data, I did the debugger and found an infinite loop in this iter() , there is only one element but the 'for' never ends. To solve this problem I changed iter() to the getchildren() function and then the process ended correctly.

felipehertzer:
Thank you for your feedback. Our system only works with articles in English for now. We will help with the correction and improvement of the package.

trafilatura/core.py
adbar: Thanks for the explanation. As a sidenote, maybe this would be necessary for all cases in `core.py` where element children are being iterated on.

trafilatura/metadata.py
```diff
@@ -160,13 +161,13 @@ def examine_meta(tree):
             if elem.get('property') == 'article:tag':
                 tags.append(content_attr)
             elif elem.get('property') in ('author', 'article:author'):
-                author = author or content_attr
+                author = content_attr if author is None else author + '; ' + content_attr if author not in content_attr and not content_attr.startswith('http') else author
```
adbar, line 164: Hi @felipehertzer, I'm sorry but the chained conditions are hard to read and to maintain, could you please rewrite the lines 164, 170, and 191 in a more standard fashion?
Simple if/else loops with a statement per line would work exactly the same and be much easier to understand.

adbar:
Thanks @felipehertzer for your further work on this, apart from the code readability issue everything looks good!

trafilatura/metadata.py
felipehertzer: Thanks for the feedback I wasn't sure which mode to use.

adbar:
@felipehertzer The `getchildren` move is rather experimental, that's why the tests fail.
Let's move it out of this PR and first complete the process.

felipehertzer:
@adbar Yes, sure, I just rollback the process.

felipehertzer:
I found this error because I'm using `include_formatting=True,`

PR #95, opened 2021-07-14
Defined a pattern to separate authors always with a semicolon

felipehertzer:
In this pull request I defined a pattern for multiple authors, where the code should always return the authors separated by a semicolon, making it easier to use the information. For this initially I defined some symbols (; | , &) and the word 'and' or 'und' (German) to separate the authors.

Before the fix `author: "Jenny Smith and John Smith"`
After the fix `author: "Jenny Smith; John Smith"`

Example where we can find multiples authors.

[SMH](https://bit.ly/3xG6U1Y)

[Nine](https://bit.ly/3i67FuO)


felipehertzer:
[Preventionaus](https://bit.ly/3hCLv4f)
I moved the trim above the check as it was returning a string with the author's name, but with too many blank characters and it wasn't passing the character limit check.

Before this fix: `"author": "By"`
After this fix: `"author": "Nina Bahadur"`

felipehertzer:
I added filter to remove some specials characters ':()?*$#!' from author names. 
In this case the author put a smile on their name.
[Pickr](https://bit.ly/36AQKv2)

Before this fix `"authors": "Leigh :) Stark"`
After this fix `"authors": "Leigh Stark"`

felipehertzer:
I found some websites/frameworks that display classes, ids, itemprops html in camel/title/uppercase, after performing the debug I found that xpath is case sensitive. In this fix I added the translate method which converts uppercase to lowercase.

`<span class="GalleryByLine__AuthorName-c00sr7-1 lnZqdp">By Author</span>`

[Example](https://bit.ly/3yYQ66D)
Before this fix: `"author": null`
After this fix: `"author":"Eden Faithfull"`

trafilatura/metadata.py
```diff
@@ -46,13 +46,26 @@ def normalize_json(inputstring):
     return trim(inputstring)
 
 
+def normalize_authors(author):
+    # fix to code with unicode
+    if '\\u' in author:
+        author = author.encode().decode('unicode_escape')
+    # simple filters for German and English
+    author = re.sub(r'^([a-zäöüß]+(ed|t))? ?(by|von) ', '', author, flags=re.IGNORECASE)
```
adbar, line 54: Yes, it's more logical to move the regexes here. It would be even better to pre-compile them above, small performance gains can be expected from moving regex compilation out of the loops.

trafilatura/metadata.py
```diff
@@ -194,10 +211,12 @@ def examine_meta(tree):
                 tags.append(content_attr)
         elif 'itemprop' in elem.attrib:
             if elem.get('itemprop') == 'author':
-                if author is None:
-                    author = content_attr
-                elif author not in content_attr and not content_attr.startswith('http'):
-                    author = author + '; ' + content_attr
+                authors = normalize_authors(content_attr).split(';')
```
adbar, line 214: It also didn't occur to me that the same code is repeated thrice. It could be moved to `normalize_authors` or at least to a separate function?

trafilatura/metaxpaths.py
```diff
@@ -5,48 +5,65 @@
...
-    """//div[starts-with(@class, 'post-info') or starts-with(@class, 'postinfo') or
-    starts-with(@class, 'post-meta') or starts-with(@class, 'postmeta') or
-    starts-with(@class, 'meta') or starts-with(@class, 'entry-meta') or starts-with(@class, 'entry-info') or
-    starts-with(@class, 'entry-utility') or starts-with(@id, 'postpath')]//a""",
-    "//p[starts-with(@class, 'postmeta') or starts-with(@class, 'entry-categories') or @class='postinfo' or @id='filedunder']//a",
-    "//footer[starts-with(@class, 'entry-meta') or starts-with(@class, 'entry-footer')]//a",
-    '//*[(self::li or self::span)][@class="post-category" or @class="postcategory" or @class="entry-category"]//a',
-    '//header[@class="entry-header"]//a',
-    '//div[@class="row" or @class="tags"]//a',
+    '//div[starts-with(' + attr_types['class'] + ', "post-info") or starts-with(' + attr_types[
```
adbar, line 27: This PR is about authors, I'd rather leave the three other XPath changes to another PR.

adbar:
Hi @felipehertzer, thanks for your input, please check my code comments.

Is performance affected by the more precise XPath expressions?

As a side note, we're reaching the limits of the LXML+XPath approach here... It's fine for now, but switching to another parser in the future could be an option. I haven't found any suitable candidate so far though.

felipehertzer:
Hi @adbar thank you very much for your feedback, this reorganization and removal of the duplicate code really made a lot more sense. I'm going to move xpath to another pull request so I can measure the performance.

felipehertzer:
I added one more fix, [this](https://bit.ly/3re3eCg) website has a json code that contains only the person tag, but the code is looking for only author.

felipehertzer:
I added a little xpath check for ids I believe many sites use id.

[Example](https://bit.ly/3BbK4Sc)
Before this fix: `"author": null`
After this fix: `"author": "Shannen Findlay"`

adbar:
Nice job, thanks!

Issue #92, opened 2021-07-12
Issue with multiple authors and preference for meta information


felipehertzer:
**We shouldnt believe on schema person**

[agenda](https://www.australianageingagenda.com.au/contributors/opinion/better-cooperation-between-providers-is-vital-opinion/) 
Current: `"author": "Sandy Cheu",`
Should be: `"author": "Stephen Teulan; Nikita Weikhardt",`

[aged](https://www.agedcareguide.com.au/talking-aged-care/only-half-of-aged-care-workers-vaccinated-as-mandatory-vaccine-deadline-approaches?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+talking-aged-care+%28Talking+Aged+Care%29) 
Current: `"author":"Consumers",`
Should be: `"author": "Liz Alderslade",`

**meta remove single names**
[cath](https://cathnews.com/cathnews/41649-north-queensland-s-wizards-of-oz-hit-the-stage) 
Current: `"author": null,`
Should be: `"author": "Rebecca",`

[echo](https://echo-news.com.au/a-delightful-dictionary-for-bowls/) 
Current: `"author": null,`
Should be: `"author": "Katie",`

adbar:
Hi @felipehertzer, thank you for your feedback! Let's examine your pull request #91 first and then move on to to this list.

adbar:
@felipehertzer Did you solve part of the problems above in the PR? Do you have other examples where author extraction fails?

felipehertzer:
@adbar These sites do not contain meta tag or the html is confusing to extract via xpath. But they use a javascript that I believe is a framework to fill in the data on the front end.

Perthnow -> `"byline":{"text":"Finn McHugh"}`
This code is found inside a simple script tag.

ESPN -> `"articles":[{"id":31807952,"author":"Andrew McGlashan","trackingName":"&lpos=:31807952"},`
This code is found inside a simple script tag.
I'm not sure, but maybe we can search for these tags in javascript


Discovery Channel -> `"author":"Ian Shive"`
This code is found inside an application/ld+json script tag but the author format is different from what we have in [REGEX](https://github.com/adbar/trafilatura/blob/7a4d16033bc496febaedb74baa4aa7e7d47b3cbb/trafilatura/metadata.py#L27)
I tried to solve the problem by adding the code `"author?".+?"([^"]+)` along with the request, but this ends up creating a new group and the current code only expects group 1

[...]

PR #101, opened 2021-07-26
Added initial Json LD parser support

felipehertzer:
In this fix i added a json parser based on schema.org, if the json is invalid there is the old parser with regex to try to capture the information.

known issue:
On some sites listed in #92, the author's name in the person schema is not the true one, where the correct thing would be to search with xpaths.

A few example of fixes:
[broadsheet](https://bit.ly/36XkPVQ) 
Before this fix: `"author": "Jake Roden",`
After this fix: `"author": "TOMAS TELEGRAMMA",`

[first](https://bit.ly/3x5yqVu) 
Before this fix: `"author": "Nick Griffin",`
After this fix: `"author": "Stan Shamu",`

[metacritic](https://bit.ly/373b08E) 
Before this fix: `"author": null,`
After this fix: `"author": "Keith Kimbell; Metacritic Film Editor",`

[thewest and more 12 sites](https://bit.ly/3zCmYlU) 
Before this fix: `"author": "The west",`
After this fix: `"author": "Steve Jancetic",`

- I also added an entry in tagsxpath and another to avoid fetching the text in the wrong place.
- I also added the author separation by slash (/)
- The metadata file was getting big, I moved the json functions to the json.py file
- I added (written by and words by) to the prefix regex
- Correction in the search for tags where the url could only contain the word 'tags' in the plural, but some frameworks use only 'tag'


adbar:
Hi @felipehertzer, I did some changes as code linting was necessary:
- The local module "json.py" was conflicting with the general one,
- There was a circular import on the `normalize_authors` function,
- I also cleaned the code a bit.

I think we can start from this version and move towards a merge. The metadata tests are failing though, I'm not sure why.

Issue #98, opened 2021-07-21
[trafilatura.core] INFO: discarding data for url: None


phongtnit:
Hi,

These urls cannot extract content because of `[trafilatura.core] INFO: discarding data for url: None`

Tested on urls:
```
https://www.biblestudy.org/bibleref/meaning-of-numbers-in-bible/17.html
https://www.biblestudytools.com/revelation/17.html
https://affinitynumerology.com/number-meanings/number-1-meaning.php
```

Sample code:
```
url = 'https://www.biblestudy.org/bibleref/meaning-of-numbers-in-bible/17.html'

downloaded = trafilatura.fetch_url(url)
result = bare_extraction(downloaded, output_format='xml', no_fallback=True, include_tables=True, include_images=True,
                         include_formatting=True, with_metadata=True, include_links=True)

if result and result['body'] is not None:
    print(etree.tostring(result['body'], pretty_print=True))
```

adbar:
Hi @phongtnit, please refer to #75, and please note that `with_metadata` is ambiguous, it will be replaced by `only_with_metadata` in the future.

phongtnit:
Hi @adbar thanks for your info, I added `url` param on `bare_extraction` function but there is the same error.

Sample code:
```
url = 'https://www.biblestudy.org/bibleref/meaning-of-numbers-in-bible/17.html'

downloaded = trafilatura.fetch_url(url)
result = bare_extraction(downloaded, output_format='xml', no_fallback=True, include_tables=True, include_images=True,
                         include_formatting=True, only_with_metadata=True, include_links=True, url=url)

if result and result['body'] is not None:
    print(etree.tostring(result['body'], pretty_print=True))
    print('\n\r')
```

adbar:
Hi @phongtnit, I see, it's because the date extraction fails for this webpage.

`only_with_metadata` specifies that title, URL and date have to be present, if one of them is missing (here the date) then nothing is returned.
If you set the parameter to `False` you'll get an output.

adbar:
Hi @phongtnit, I believe the problem has been addressed by my answer and by recent commits. If you still run into a bug you can re-open the issue.

Issue #96, opened 2021-07-15
Articles without date, or with date without id, class or props


felipehertzer:
Is the date really mandatory to extract the file? can we create a regex to fetch dates in html?

Here is an example where the article has a date but it is difficult to find it without the regex.
[Example](https://www.cis.org.au/commentary/articles/the-gilded-cage-down-under/)

adbar:
@felipehertzer The date is only necessary if the parameter `only_with_metadata` is set to `True`.

In your example I would say that htmldate's logic is to find several occurrences of `/2021/07/` (here in URLs) and deduce that the article is from July 2021.
Pushing the granularity to the very day (the 9th) doesn't seem doable with a generic approach, or only at the cost of looking for regexes in the whole page, which comes at a computational cost. Nonetheless, it's what htmldate's extensive mode is about.

felipehertzer:
After correcting the date search, it is now working perfectly. Thanks.

PR #106, opened 2021-07-29
Fix only_with_metadata site without URL or Date

felipehertzer:
Fix in only_with_metadata condition reported in issues: #96 #98 #92

Added support to [parsely tags](https://github.com/Parsely/wp-parsely/blob/8cc3b5f8df77e3e2fc3fe614e9774aa0938904d2/views/repeated-metas.php)

Fixed author names when is one name should not be big


trafilatura/core.py
```diff
@@ -747,7 +747,7 @@ def bare_extraction(filecontent, url=None, no_fallback=False,
                 LOGGER.info('blacklisted URL: %s', url)
                 raise ValueError
             # cut short if core elements are missing
-            if only_with_metadata is True or with_metadata is True and any(
```
adbar, line 750: `with_metadata` isn't deprecated yet. There could be a line `only_with_metadata = with_metadata` earlier in the code though.

trafilatura/utils.py
```diff
@@ -322,12 +322,14 @@ def normalize_authors(current_authors, author_string):
         author = AUTHOR_REMOVE_NUMBERS.sub('', author)
         author = AUTHOR_REMOVE_PREPOSITION.sub('', author)
         # skip empty strings
-        if len(author) == 0:
+        if len(author) == 0 or (' ' not in author and len(author) >= 23):
```
adbar, line 325: Is there a good reason to set 23 as threshold? What about other languages than English?

trafilatura/core.py
adbar: @felipehertzer I will take care of that.

trafilatura/utils.py
felipehertzer: The average of the first name is 6 characters, for english and portuguese.  I believe maybe German first name have a few more letters.  To set a number I used this ISO https://en.m.wikipedia.org/wiki/ISO/IEC_7813

trafilatura/core.py
felipehertzer: I had problems with cli that was not passing the variable with_metadata to the extractor() and this error should only have been called when only_with_metadata is required.  Alrgiht I'll leave it to you

adbar: I think I fixed the CLI problems in 3eafd7e5c09c7b49c513342eb5089544f29c189e, if something still isn't working we can add tests for it.

trafilatura/utils.py
adbar: Thanks for the reference! It says 26 characters including separators, but there are also composite names with a hyphen. It would be a pity to miss information by being too restrictive here.

trafilatura/core.py
adbar: Line added in 968721971c0d65f9fe95b983ca050c0f97d4b926.

trafilatura/utils.py
adbar: Constraint relaxed in 8eb2605.

adbar:
@felipehertzer Everything OK? Ready to merge?

felipehertzer:
Hi @adbar looks good for me!

Issue #53, opened 2021-01-12
Bypass catchas/cookies/consent windows?


adbar:
Are there Python libraries which allow to bypass diverse consent mechanisms put in place by news outlets for the readers to allow cookies? It would be too cumbersome to develop a headless browser exclusively for trafilatura.

A good example would be the newspaper zeit.de. Related to #18.

Potential solutions:
- headless browser with automatic clickling mechanism
- use [AMP](https://www.howtogeek.com/284166/what-is-google-amp-and-why-is-it-in-my-search-results/)-links

The output could be piped directly to trafilatura (in a terminal or via Python).

adbar:
For interaction with websites also see [pypetteer](https://github.com/pyppeteer/pyppeteer), a headless chrome/chromium automation library (unofficial port of puppeteer).

adbar:
This is a possible way to use cookies: https://github.com/adbar/trafilatura/pull/108#issuecomment-908260901

adbar:
This goes beyond the scope of the software, closing the issue for now.

PR #108, opened 2021-08-08
Add setting for cookies in request headers

muellermartin:
This PR adds the ability for setting cookies in request headers. Cookies may sometimes be required to access pages behind a login or some kind of consent page.

Might be a partial solution for #53.

Note: This solution works for zeit.de by using something like `config['DEFAULT']['COOKIES'] = 'zonconsent=' + datetime.now().isoformat()` for the cookie header.

### DIFF
--- a/tests/downloads_tests.py
+++ b/tests/downloads_tests.py
@@ -13,7 +13,7 @@ from unittest.mock import patch
 from trafilatura.cli import parse_args
 from trafilatura.cli_utils import download_queue_processing, url_processing_pipeline
 from trafilatura.core import extract
-from trafilatura.downloads import add_to_compressed_dict, fetch_url, decode_response, draw_backoff_url, load_download_buffer, _determine_headers, _handle_response, _parse_config, _send_request
+from trafilatura.downloads import USER_AGENT, add_to_compressed_dict, fetch_url, decode_response, draw_backoff_url, load_download_buffer, _determine_headers, _handle_response, _parse_config, _send_request
 from trafilatura.settings import DEFAULT_CONFIG, use_config
 from trafilatura.utils import load_html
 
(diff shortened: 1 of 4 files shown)


muellermartin:
Hi, thanks for the changes! Why is the `COOKIES` option renamed to `COOKIE`? Is there a problem with multiple cookies? I saw that older versions of Python treated an inline semicolon as start of a comment, but in recent versions of Python this seems to be fixed, or is there another reason?

adbar:
Hi @muellermartin, thanks!

We could also think about supporting a series of cookies if you're interested – probably as key-value pairs _website/cookie_.

muellermartin:
Thanks for merging and your additions <3

As far as I understand it, multiple cookies can now be added by separating them with a semicolon in the `Cookie`header ([reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie)). A better interface for adding cookies per website like suggested would be nice though. I thought about using something like a [CookieJar](https://docs.python.org/3/library/http.cookiejar.html), but the currently used urllib3 seems not able to use them...

adbar:
It seems feasible via a `SessionManager` but I cannot find a user guide for it in the docs:
- https://github.com/urllib3/urllib3/pull/904
- https://urllib3.readthedocs.io/en/latest/changelog.html?highlight=cookiejar#id24

Foorack:
> 
> 
> It seems feasible via a `SessionManager` but I cannot find a user guide for it in the docs:
> 
>     * [Approved SessionManager changes urllib3/urllib3#904](https://github.com/urllib3/urllib3/pull/904)
> 
>     * https://urllib3.readthedocs.io/en/latest/changelog.html?highlight=cookiejar#id24

It seems like 904 was merged into "urllib3:sessionmanager", but this was then deleted, so it never got into master. It seems like the idea is to use CookieJar instead...

https://github.com/urllib3/urllib3/pull/1229/files#diff-8bb5645ed9da6ed97491165b18352e739650f1d4422e2134c6f5500d6384b6daR39
and
https://docs.python.org/3/library/http.cookiejar.html#http.cookiejar.CookieJar.add_cookie_header

adbar:
@Foorack thanks!
Would any of you be interested in trying this in a PR? (cc @muellermartin)

PR #110, opened 2021-08-18
Added support to author blacklist and discard elements

felipehertzer:
In this fix two important changes were added, the first one we can now remove parts of the site that can make extraction difficult, for example (sidebar, comments). Also, the option to add names to be ignored by the project was added. Like generic names in meta and json.

`result = extract(html, url=url,
                 with_metadata=True,
                 output_format='json',
                 author_blacklist={'Watermark Feed'})`

In this fix:
- Added check to avoid email on author name
- Fixed wrong check on util/normalize_authors and &amp;
- Added discard xpath to authors
- Added tags on metaxpath
- Added one more prefix
- Added tag on comments xpath (Metacritic.com)
- Added support to author_blacklist to prevent fake or wrong authors in meta/json/source

Example:
[Aap](https://bit.ly/3yYlN0o) 
Before this fix: `"author": "Watermark feed",`
After this fix: `"author": "Aaron Bunch",`

[angelicans](https://bit.ly/3yZMqlo) 
Before this fix: `"author": "Sydneyanglicans net",`
After this fix: `"author": "Hannah Thiem",`

[Armida](https://bit.ly/37PWHEP) 
Before this fix: `"author": "Bob Freebairn",`
After this fix: `"author": "Mark Griggs",`

[Augamers](https://bit.ly/37PXK7Q) 
Before this fix: `"author": "Sean",`
After this fix: `"author": "KostaAndreadis",`

[sportline](https://bit.ly/3mcCn9d) 
Before this fix: `"author": "Justin Perri Model Guy",`
After this fix: `"author": "Justin Perri",`

[abc](https://ab.co/3jX3mCY) 
Before this fix: `"author": "Getty Images",`
After this fix: `"author": null,`

[metacritic](https://bit.ly/3spuFti) 
Before this fix: `"author": "ni554n",`
After this fix: `"author": "Jason Dietz; Metacritic Features Editor",`

The issue #92 was updated!


adbar:
@felipehertzer Thanks for the PR, it looks good, I'll be AFK for some time and will get a closer look at it when I get a chance.

trafilatura/xpaths.py
```diff
@@ -36,7 +36,7 @@
     contains(@class, 'single-post') or
     contains(@class, 'main-column') or contains(@class, 'wpb_text_column') or
     starts-with(@id, 'primary') or starts-with(@class, 'article ') or @class="text" or
-    @class="cell" or @id="story" or @class="story" or
+    @class="cell" or @id="story" or @class="story" or contains(@class, "body-description") or
```
adbar, line 39: Not necessary as per https://github.com/adbar/trafilatura/issues/112#issuecomment-919099605

Each expression slows things down a bit and there is a powerful algorithm working as fallback so I would refrain from adding to many constraints to `xpaths.py`.

felipehertzer:
HI @adbar  sorry for the delay, alright I did the rollback, I'll avoid adding new elements on xpath, but probably in the future we will have to find a way to remove the elements more efficiently.

adbar:
Hi @felipehertzer, I checked the code and made a few commits. Could you please address the remaining questions above?

For now I see one XPath expression to decide on, I suggest we keep track of them in a separate issue, feel free to add further ones here in the future: https://github.com/adbar/trafilatura/issues/115.

felipehertzer:
Hi @adbar , I can't see any more questions and comments, what would the questions be?

adbar:
@felipehertzer I entered a few comments but something seems to have changed on GitHub and I somehow started a review... Maybe you can't see it above, here are the questions.

Screenshot_2021-09-20_11-19-59

felipehertzer:
Hi @adbar sorry I was busy on another project.

In the first question I added prune to remove all elements that might contain tags like those used in author, such as author of a comment, or a list in the sidebar with the most read articles. So before performing the search I perform this command to avoid searching for component tags that can interfere with the search, besides the extract_author function only searches the first author with the most specific tag.

In the second question I performed the second check because we only perform the search in HTML if we don't find any author in meta and/or json, but sometimes we have to block something in json to then perform the search in html and if even the html brings incorrect then I thought it was better to recheck the blacklist to avoid bringing the author that should be removed.

I really haven't found a better way to avoid searching for author tags inside components like comments or sidebars, would you have any suggestions to improve it?

Thank you.

adbar:
Hi @felipehertzer, thanks for the answers. Yes, I guess there is no other easy way to perform the task. I going to perform a few checks and see if I can merge the PR.

adbar:
Hi @felipehertzer, according to my benchmark the code got slower between version 0.9 and 0.9.1 but there is nothing worth mentioning with this change, so I will merge it.
Please note that we should analyze code performance at some point to see if there are easy gains to be made in metadata extraction. Otherwise my guess is that the package is going to get more and more precise over time but also slower.

PR #130, opened 2021-10-31
Minor optimizations and fixes

vbarbaresi:
### Description

I don't expect a major performance improvement from these little changes, these are small nits.

To get a major improvement, I think iterating on documents in a single-pass (using `etree.iterparse` instead of using lots of `etree.xpath`) would be the way to go

### Detais

- add `utils.unifiqy_list()` to factor `list(OrderedDict.from_keys())`
It turns out that we don't need OrderedDict anymore for this since Python 3.6 and it's much faster with a simple `dict`

- use generators in `extract_comments()` instead of building lists

- use sets instead of lists for `MANUALLY_STRIPPED` and `MANUALLY_CLEANED`
  using ` .remove()` on a list is slow (`O(n)`) so a set seemed more appropriate. It probably doesn't make a difference given the size of these sequences though..

- pass `*seq` in `etree.strip_tags(tree, seq)`
  It seems that it works with a list too but it shouldn't according to the method signature

### Additional notes

Note on uniqify benchmark: I used this simple benchmark:

```python
import random
import time

seq = [random.randrange(0,100000) for i in range(1000000)]

t = time.time()
dedup = list(dict.fromkeys(seq))
print(time.time() - t)
```

to compare OrderedDict and Dict
and also tried using lists with random strings inspired by this code: https://gist.github.com/peterbe/67b9e40af60a1d5bcb1cfb4b2937b088

### DIFF
--- a/trafilatura/cli_utils.py
+++ b/trafilatura/cli_utils.py
@@ -33,7 +33,7 @@ from courlan import get_host_and_path, is_navigation_page, validate_url
 from .core import extract
 from .downloads import add_to_compressed_dict, buffered_downloads, load_download_buffer
 from .filters import content_fingerprint
-from .utils import uniqify_list
+from .utils import uniquify_list
 from .settings import (use_config, FILENAME_LEN,
                        FILE_PROCESSING_CORES, MAX_FILES_PER_DIRECTORY)
 from .spider import get_crawl_delay, init_crawl, process_response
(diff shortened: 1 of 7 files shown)


vbarbaresi:
This Sourcery bot is a quite invasive! Re-opening a PR with my code 😅

adbar:
Yes, the Sourcery bot was too much, I switched it off and will now close PR #131.

adbar:
Thanks!
No changes in speed or accuracy on my data, although deletion order of HTML elements could matter.
I edit the PR to accept changes and rollback others.

adbar:
@vbarbaresi please give me `push` access to the repository, I need to commit a few more changes.

PR #152, opened 2021-12-18
Bugfix when articleSection is null - json_metadata

felipehertzer:
I found some sites that in articleSection the value is null and it was breaking the package.

Example:
```json
"articleSection": [
    null
],
```

Solution: Remove any null from articleSection tag


adbar:
Hi @felipehertzer, that looks good, thanks!

PR #186, opened 2022-03-10
Navigation bar filtering - some bug fixed

immortal-autumn:
The current repo should work well? I have removed several things that are unused and fixed a tiny bug that affects the accuracy. I have added to the git ignore so that the branch should now get quite clean as well XD


immortal-autumn:
Yea I found that, but I think that two failed tests leads by the modification of the original webpage.

The failed test here:
` result = load_mock_page('https://www.stuttgart.de/item/show/132240/1', xmloutput)
    assert 'Das Bohnenviertel entstand' in result and 'sich herrlich entspannen.' in result and 'Nützliche Links' not in result # and 'Mehr zum Thema' not in result`
has a page-not-found exception, which fails the test.

For the modification of the webpage, I uses the original source from the current branch recovers the whole test cache. I loaded the differences and found that they are actually the same.

I will remove the .idea for sure.

adbar:
@immortal-autumn Listing `.idea/` in the `.gitignore` file is a good idea, but you also have to remove your `.idea/*` files from this PR.

The tests still don't pass, which means your PR leads to a regression on this data, not to an improvement...

adbar:
Thanks for sharing the dataset, I will have a look at it. Chinese websites are quite different from Western ones and (sadly) many heuristics cannot be applied indifferently to the ones or the others.

Your approach does not integrate well with my current tests and my current benchmark, that's why there are errors. For things to work we need to clean your code branch, that's still quite a lot of work.

adbar:
@immortal-autumn 等一下 ;)
I understand you don't work to work actively on the PR. Yet there are easily integrable ideas like `WeChat|QQ|WeiBo`, I'm going to see what I can salvage.

adbar:
It's cleaner, now we need to test and think about what the changes bring in terms of improvements or regressions.

adbar:
The commit [b15085f](https://github.com/adbar/trafilatura/pull/186/commits/b15085ff3de700d7bf9ba0c7605f9b6cd8b7b64b) references a working version of the PR, albeit slower.

From there I don't see what the `remove_nav()` function brings, it markably degrades performance on my data, moreover there is already a link-based filtered for `div` sections:
https://github.com/adbar/trafilatura/blob/e975f1bce169da25f3e7d8ca00739109592b15d3/trafilatura/htmlprocessing.py#L117

adbar:
The PR now looks ready to be merged, some of the changes you suggested could be kept.

Any comments @immortal-autumn?

immortal-autumn:
@adbar oh I found what happened to why some navigation bars are not filtered in some of the websites, which leads by the difference between Chinese and English characters in the link density filter. I must have some misunderstandings about the function, I will continue to investigate that XD.

adbar:
I'm re-opening the PR since I want to merge it!

Yes, the algorithms rely on heuristics related to text length, Chinese websites are different in this regard. You can also try to adjust the values in `settings.py`.

immortal-autumn:
Oh sorry about that lol, I just not familiar with PR.

PR #188, opened 2022-03-16
Update xpaths.py

glacierck:
from http://news.china.com.cn/


adbar:
@glacierck Thanks for the suggestion! I don't understand the error above, everything should work properly. I'll look at it later.

adbar:
Everything looks OK, I'll merge this change after PR #186.

PR #192, opened 2022-04-08
Fixed bug with liveblogposting when is missing liveblogupdate

felipehertzer:
Hey @adbar.

I'm opening this pull request to fix liveblogposting. Some websites only show liveblogupdate when the event starts. It was throwing an exception.

Thanks.


adbar:
Hi @felipehertzer, the PR looks good, thanks!

PR #197, opened 2022-04-14
Added Coinbase article annotation

swetepete:
Annotation is in evaldata.py, html page source in /eval


adbar:
@swetepete are you still working on this or should we consider merging/aborting?

adbar:
I could salvage a few annotated files for the evaluation, that's better than nothing. The rest wasn't accessible or easy to reproduce (AWS account necessary etc.)

PR #198, opened 2022-04-18
Encoding and blank list filtering

immortal-autumn:
I have added some logic to xml.py and utils.py, which helps with filtering unusual encodings and blank tables. This could help with improving the accuracy of the application.


adbar:
Hi @immortal-autumn, apparently you didn't use the latest version of the repository or didn't perform a merge before sending your PR. This makes your actual changes difficult to spot...

- If I understand correctly you added two regexes in order to better filter the output?
- Can it be that you added an evaluation file but there is no corresponding evaluation?

immortal-autumn:
Oh the evaluation file I added is for the self-evaluation use, I forgot to remove that sorry lol
The modification happens in the last three commits, I am actually not sure about why there are a number of commits here.

adbar:
Hi @immortal-autumn, thanks for the feedback, I'm going to have a look at the potential improvements when I find the time.

adbar:
Hi @immortal-autumn, now I looked at it and I don't understand: The characters you wish to remove are already accounted for. The ideographic space character in particular (`\u3000`) is already handled here:
https://github.com/adbar/trafilatura/blob/146506a2a18e5ca99ee8cd9a5779cc9f137697aa/trafilatura/utils.py#L44

Thus I don't have the impression that your PR brings anything. Unless you can name a case in which extraction improves with it, I will discard the PR.

Issue #195, opened 2022-04-12
Extend test coverage for json_metadata functions


adbar:
Hi @felipehertzer, you have contributed significant parts of the code in `json_metadata.py`. As you know they are error-prone as certain web pages don't hold to standards. We can't do much about this code, but we can work on the code coverage. It is [rather low](https://app.codecov.io/gh/adbar/trafilatura/blob/146506a2a18e5ca99ee8cd9a5779cc9f137697aa/trafilatura/json_metadata.py).

Improving coverage would allow us to be sure the code works as intended and all potential errors are caught. Could you please help with this?

felipehertzer:
Hey, sure I will check it ASAP.

adbar:
Hi @felipehertzer, there are still untested lines, see [online coverage report](https://app.codecov.io/gh/adbar/trafilatura/blob/20dba1988949f7caaa2ef80dc01350d31f8acf35/trafilatura/json_metadata.py) (you may have to try visiting the URL at a different time to see the report).

Since there are regularly issues related to this part of the extractor you may want to increase coverage.

felipehertzer:
Hi @adbar, sure I will check it.

adbar:
For further reference, here is a new URL to follow the coverage:
https://app.codecov.io/gh/adbar/trafilatura/blobs/master/trafilatura/json_metadata.py

adbar:
It changed again, here is the new URL, we're nearly there in terms of coverage:
https://app.codecov.io/gh/adbar/trafilatura/blob/master/trafilatura%2Fjson_metadata.py

PR #204, opened 2022-05-03
Fixed multiple spaces on author name and added more ld json tests

felipehertzer:
Hey @adbar this is the first part of #195, I added some examples and fixed the authors names, I was creating names with multiple spaces between the names


adbar:
Hi @felipehertzer, that looks good, thanks!

PR #209, opened 2022-05-18
Added 'article:author' as meta name

felipehertzer:
Hey @adbar I found a case where some sites are using the article:author property as meta name and the code was unable to find the correct names in some cases.

I saw a few Wordpress websites that are using that tag incorrectly. Probably a plugin or some theme with this bug.

example:
Couriermail
https://bit.ly/3MuDGus

```
Current: Lloyd Clarke Hannah Clarke's father
Should be: Lloyd Clarke
```


adbar:
Thanks @felipehertzer! I found other interesting attributes (apparently: deprecated Facebook guidelines) and added them.

felipehertzer:
That's makes sense. Thanks!

PR #213, opened 2022-05-27
Fixed Cannonical URL with a path, wrong author name with a quote, wrong author name with author-meta

felipehertzer:
Hey @adbar 

I fixed some issues that I found during testing. Also I add the prefix 'from' in normalize authors, and I add the class meta-date and meta-review on authors xpath.

**The cannonical starts with a path and it should be an URL, I just changed the match to accept only url.**
`<link rel="canonical" href="/article/00139....." />`

> Example: [idm](https://bit.ly/3sYKubS)
> Current: `None`,
> Should be: `"title": "Tasmania’s $A150....", "author": null, "hostname": ,`
> 
> Example: [sen](https://bit.ly/38n3lX5)
> Current: `None`,
> Should be: `{"title": "Collingwood defende..", "author": null, "hos`

**Second fix is to avoid quote author and get the article author:**

> Example:  [mi-3](https://bit.ly/38qXQXA)
> Current: `— Patrick Carne; Co-Founder; CEO; Inlight`,


adbar:
Hi @felipehertzer, thanks! The PR looks good but how do you explain the regression in terms of code coverage?

adbar:
@felipehertzer please not that I still don't understand the lower code coverage, but ok.

felipehertzer:
Hi @adbar, sorry I missed that comment. Maybe the changes on metaxpath, but I not sure why.

adbar:
@felipehertzer No worries. I think it was because of relative URLs but they way they are being handled must have changed in-between. Anyway, let's keep an eye on this.

Issue #226, opened 2022-07-24
Trafilatura exception with specific URLS


dmoklaf:
Some rare URLs trigger the follow stack trace (below).

Environment:
MacOS 12.4
Python 3.10.5
Trafilatura 1.2.2

```
could not extract web page content: https://a16z.com/author/ben-horowitz/?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed:+bhorowitz+(ben's+blog)
Traceback (most recent call last):
  File ...
    extraction = trafilatura.core.bare_extraction(
  File "/usr/local/Caskroom/mambaforge/base/lib/python3.10/site-packages/trafilatura/core.py", line 813, in bare_extraction
    document = extract_metadata(tree, url, date_extraction_params, no_fallback, author_blacklist)
  File "/usr/local/Caskroom/mambaforge/base/lib/python3.10/site-packages/trafilatura/metadata.py", line 435, in extract_metadata
    metadata = extract_meta_json(tree, metadata)
  File "/usr/local/Caskroom/mambaforge/base/lib/python3.10/site-packages/trafilatura/metadata.py", line 107, in extract_meta_json
    metadata = extract_json(schema, metadata)
  File "/usr/local/Caskroom/mambaforge/base/lib/python3.10/site-packages/trafilatura/json_metadata.py", line 60, in extract_json
    if 'name' in content and not content['name'].startswith('http'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

felipehertzer:
Hi @adbar, you can assign it to me. I will fix it tomorrow morning here.

PR #227, opened 2022-07-26
Fixed bug where author name could be null

felipehertzer:
Hey @adbar,

The fix for the bug reported in #226


adbar:
LGTM, thanks @felipehertzer!

PR #235, opened 2022-08-12
Extend test coverage for json_metadata functions, added fallback to author name when name is a string on article type

felipehertzer:
Hey @adbar,

I'm sending more tests to json metadata functions.

Thanks.

#195

### DIFF
--- a/tests/json_metadata_tests.py
+++ b/tests/json_metadata_tests.py
@@ -525,6 +525,56 @@ def test_json_extraction():
 </body></html>'''), metadata)
     assert metadata is not None and metadata.title == "12 words and phrases you need to survive in Hamburg" and metadata.author == "Alexander Johnstone" and metadata.sitename == "The Local"
 
+    metadata = Document()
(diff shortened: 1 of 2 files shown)


adbar:
Thanks @felipehertzer! I don't understand the coverage diff report, everything looks fine.

Did you mark the PR as draft? Are you still working on it?

adbar:
I just merged it since everything looked OK.

felipehertzer:
Hi @adbar,

I was checking the read json function https://github.com/felipehertzer/trafilatura/blob/fdab3df410f5cb4946f404ae305c41844a1b740e/trafilatura/metadata.py#L104

The regex is to minify the json but it is removing the ```\"``` when a value contains double quoutes ```"this is \"test2\" a test"```. Some tests are falling into except. But I could find a better way to minify and avoid to remove the escapes.

The coverage problem is that we need to test the json reading correctly and the json with problem.

adbar:
Hi @felipehertzer, thanks for the info, feel free to make another PR if you make any progress!

PR #238, opened 2022-08-30
TEI: Add <publisher> if license is added

knit-bee:
Changes:
- I moved the extraction of the `publisherstring` in the `write_full_header` function to a separate function to be able to reuse it.
- I changed the `write_full_header` function so that a `<publisher/>` element is added before the `<availability/>` element in  `<publicationStmt/>`. The text that is used for the `<publisher/>` element is the same as in `sourceDesc/biblFull/publicationStmt/publisher`.

Background:
A document is not valid according to TEI P5 if the `fileDesc/publicationStmt` contains an `<availability/>` element but no `<publisher/>`. Therefore, the  `<publisher/>` should also be added before.

### DIFF
--- a/tests/xml_tei_tests.py
+++ b/tests/xml_tei_tests.py
@@ -1,0 +1,73 @@
+"""
+Test for transformation to TEI.
+"""
+from lxml.etree import Element
+
+from trafilatura.metadata import Document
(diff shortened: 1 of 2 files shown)


adbar:
@knit-bee Thanks! It looks good.

PR #241, opened 2022-09-03
Use "\n" for delimiter instead of "," for multiple user-agents

### LINKED ISSUE #240: Use a different delimiter for user agent setting
Hi! First of all, thank you for your amazing work. This is a great library!

I was trying to have a list of multiple user strings in order to rotate between them. However, user strings can actually contain the character `,` such as the user string: "`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/605.1.15 (KHTML, like Gecko)`". Since the library uses `split(',')`, we can get malformed user agents such as "` like Gecko)`" in the example above.

May I recommend changing how to provide multiple user agents in the settings list? I would recommend using a different delimiter (maybe `;;`) that is not used in normal user agents. I can make a quick and simple PR for this as well if you want.

k-sareen:
User-agents can contain the string "," which can lead to malformed
user-agents when using `split(",")`. This commit fixes this issue by
changing the delimiter to a newline ("\n").

Fixes #240.

### DIFF
--- a/tests/resources/newsettings.cfg
+++ b/tests/resources/newsettings.cfg
@@ -7,8 +7,13 @@ MAX_FILE_SIZE = 20000000
 MIN_FILE_SIZE = 10
 # sleep between requests
 SLEEP_TIME = 0.25
-# user-agents here: agent1,agent2,...
-USER_AGENTS = Firefox,Chrome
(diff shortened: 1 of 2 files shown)


adbar:
lgtm

PR #242, opened 2022-09-05
Set tail of element to None if tail was added as text

knit-bee:
Changes: 
- I set the tail of an element to `None` if its tail was added as text in the `process_node` function

In the `process_node` function, the tail of an element is added as text if the element doesn't contain any text but the tails stays the same, so the tail content is duplicated.
 https://github.com/adbar/trafilatura/blob/e81cffe11cb8de03cb177fd2cd3055c7d1824834/trafilatura/htmlprocessing.py#L346-L347

I think the tail should then be set to  `None` (resp. empty string), as in 
 https://github.com/adbar/trafilatura/blob/e81cffe11cb8de03cb177fd2cd3055c7d1824834/trafilatura/htmlprocessing.py#L314-L318


adbar:
Indeed, good catch!

Issue #93, opened 2021-07-12
Simplify handling of nested elements


adbar:
Processing of nested elements (within paragraphs, lists, etc.) could be simplified and optimized, the code would be shorter and easier to maintain.

Related to #60, #78, #90.

adbar:
See also #94:

It would be useful to check all cases in `core.py` where element children are being iterated on to avoid (rare) endless loops (see discussion in #91).

`element.iter()` could be replaced by `element.getchildren()` or `list(element)` if these two conditions are met:

- content order preserved
- not significantly slower

Sadly, both solutions affect content order or output.

PR #243, opened 2022-09-07
Change handling of tables

knit-bee:
Hey @adbar , I noticed that for `<cell>` elements the tag `<cell>`  is added two times if the element contains text and other elements (probably because a tag was removed during cleaning the html) 

Changes:
- I replaced `Element.iter()`  with `Element.iterdescendants()` in the `handle_table` function. `.iter` also includes the element itself, whereas `.iterdescendants` excludes it.
- I also removed the counter variable `i` and changed the `break` statement to only break if a `<table>` tag is met. Because the  the whole table element won't be processed, any `table` tag that is seen, belongs to a nested table.
- Because I used `.iterdescendants` for the children of a `td`/`cell` element as well, the cell itself is now excluded from the inner loop. So any text or tail would be lost. Therefore, I also added the text and tail of the cell to newly constructed `<cell>` element.
- I added some more tests for the `handle_table` function.

Also,  I think  `'cell'`  should be added to the set `WITH_ATTRIBUTES` in `xml.py` otherwise the information about the table header cells will be lost in the xml output.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -836,6 +836,78 @@ def test_table_processing():
     result = extract(htmlstring, no_fallback=True, output_format='xml', config=ZERO_CONFIG, include_tables=True)
     # todo: all elements are there, but output not nested
     assert '<cell role="head">1</cell>' in result and '<cell>2</cell>' in result
+    nested_table = html.fromstring(
+        """
+        <table>
+        <tr>
+        <td>
+          <table><tr><td>1</td></tr></table>
+        </td>
+        </tr>
+        </table>"""
+    )
+    processed_table = handle_table(
+        nested_table, TAG_CATALOG, dedupbool=False, config=DEFAULT_CONFIG
+    )
+    result = [
+        (el.tag, el.text) if el.text is not None and el.text.strip() else el.tag
+        for el in processed_table.iter()
+    ]
+    assert result == ["table", "row", "cell", "table", "row", ("cell", "1")]
+    complex_nested_table = html.fromstring(
+    """
+    <table>
+    <tr>
+    <td>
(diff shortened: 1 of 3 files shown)


adbar:
Thanks @knit-bee!

- `iterdescendants()` is really a good idea, it could also be used for other elements and possibly for #93 
- I just added the `cell` attribute following your remark :heavy_check_mark: 
- Since you're at it there is still an issue with nested tables which could be improved. In the following test all elements are present but the tables are not nested anymore in the output. Could you please have a look at it to see if you find a solution? It's a rare case but it happens. https://github.com/adbar/trafilatura/blob/ebd19dde893de49a05a3aea2d2625097ba622e1d/tests/unit_tests.py#L835

adbar:
The tests pass but there is an issue with benchmark: [4c4ca2b](https://github.com/adbar/trafilatura/pull/243/commits/4c4ca2bce0b0fa608bd415369d8bb947578e1ea7) degrades performance.
I tried using `iterdescendants` instead of `getchildren` but it doesn't seem to be the problem.

Could you please fix it or revert the branch to an interior state?

adbar:
@knit-bee We could revert to [ebd19dd](https://github.com/adbar/trafilatura/pull/243/commits/ebd19dde893de49a05a3aea2d2625097ba622e1d) for now and work on the rest in another PR, what do you think?

adbar:
Note: reverted after discussion, should be ready to merge.

knit-bee:
I think you need to revert commit [397b223](https://github.com/adbar/trafilatura/pull/243/commits/397b2231d75607222ccc6d6a205dde7617a59b99) also. I added some tests for nested tables in this commit that are no failing.

PR #244, opened 2022-09-08
Update quickstart.rst

dsgibbons:
The quickstart docs include a deprecated argument. PR changes no-comments example to use `output_format="xml"` as opposed to `xml_output=True`.


adbar:
Hi @dsgibbons, good catch, thanks!

PR #250, opened 2022-09-19
Handle lists

knit-bee:
Changes:
- I gathered all tests that test list processing to in a common function
- I used `iterdescendants` for processing elements in list items to avoid `<item>` as child of `<item>` for items that contain text and other elements.
- I added text content of `<list>` to a new `<item>` element to avoid text in lists that is not wrapped by an item.
- I removed `tail` on `<item>` elements (the tail is added to the `<item>` text or as tail of the last child, if present)
- I set the tag of the list to 'done' after processing (otherwise an empty list might be remain in the output)

I also had to change one of the existing tests for lists (the one for nested lists) because there the result was expected to have `<item>` as child of `<item>`. Unfortunately, the new code also changed the formatting of the result string, so I had to remove whitespace and newline chars for the assert statement to pass.

Background:
 Text in lists that is not wrapped by `<item>`,  tail on `<item>`  or `<item>` as child of `<item>` is not valid in TEI P5.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -926,18 +926,18 @@ def test_list_processing():
 </ul>
 </article></body></html>'''
     my_result = extract(htmlstring, no_fallback=True, output_format='xml', config=ZERO_CONFIG)
-    assert '''
+    expected = '''
     <list rend="ul">
       <item>Coffee</item>
-      <item>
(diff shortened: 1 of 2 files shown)


adbar:
@knit-bee Great job, LGTM.

PR #251, opened 2022-09-20
Transform <fw> with subelements to <ab>

knit-bee:
Changes:
- Handling of  `<head>` elements with children in `check_tei` function: if the `<head>` contains children, a new element with tag "ab" is constructed. Any child of `<head>` with tag "p" is removed and the text content added as tail of the previous child. If this tail is already set, a new `<lb>` element is added.

 In the current transformation to TEI,  the tag `<head>` is converted to `<fw>`. In some (probably rare cases), the `<head>` element also contains children. However, in TEI P5, it is not valid for `<fw>` to have certain subelements (e.g. `<p>`, `<list>`) . I'd suggest using `<ab>` instead of `<fw>` because it can handle more elements that might appear as children.
Unfortunately, `<ab>` also shouldn't contain any `<p>`, so in the new transformation a `<lb/>` is added if necessary.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -648,6 +648,9 @@ def test_tei():
     assert xml.write_fullheader(header, docmeta) is not None
     docmeta.title, docmeta.sitename = None, None
     assert xml.write_fullheader(header, docmeta) is not None
+    htmlstring = html.fromstring("<html><head/><body><div><h2><p>text</p></h2></div></body></html>")
+    extracted = extract(htmlstring, url='mocked', no_fallback=True, output_format="xmltei")
+    assert xml.validate_tei(etree.fromstring(extracted)) is True
 
 
 def test_htmlprocessing():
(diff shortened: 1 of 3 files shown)


adbar:
@knit-bee If I understand correctly it could be useful to remove `fw` from `TEI_VALID_TAGS` in `xml.py` once and for all? Or is `ab` used just in certain cases?

What is your take on attributes? We could pass on the title type (e.g. `h3`) as attribute, however I see no test for it in your additions.

knit-bee:
I'd suggest using `<ab>` in general for  `<head>` tags instead of `<fw>`  ( in TEI P5, [fw](https://tei-c.org/release/doc/tei-p5-doc/en/html/ref-fw.html) is for example used for header, footer or page numbering). So yes, `<fw>` could be removed from `TEI_VALID_TAGS`.
The setting of the attribute for the title type can be done as before imo. I can add a test for that.

adbar:
@knit-bee OK, let's do it!

adbar:
superseded by #254

PR #252, opened 2022-09-26
TEI output: handle tail of certain elements

knit-bee:
Changes: 
- I added some code to clean the TEI output and remove tails on `<p>`, `<fw>`, `<ab>` and `<lb/>`
   - for `<p>` elements: if they have a tail, the tail is merged with the text content
   - for `<fw>` and `<ab>`, a new `<p>` sibling is added and the former tail added as its text
   - `<lb/>` that are direct children of `<div>` are converted to a `<p>` and the tail is added as text
 
Background:
Tail on elements like `<p>` and `<fw>` usually results in a validation error with TEI P5. The same is true for `<lb/>` that have `<div>` as parent.

### DIFF
--- a/tests/xml_tei_tests.py
+++ b/tests/xml_tei_tests.py
@@ -1,10 +1,10 @@
 """
 Test for transformation to TEI.
 """
-from lxml.etree import Element
+from lxml.etree import Element, fromstring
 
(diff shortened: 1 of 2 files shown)


adbar:
superseded by #254

PR #253, opened 2022-09-27
TEI: Handle invalid siblings of <div>

knit-bee:
Changes:
- I added a check to the `check_tei` function to search siblings of `<div/>` elements for unallowed tags. If such tags (e.g. `<p/>`) appear, these elements are wrapped in an additional `<div/>` element. 

In TEI P5, some elements (e.g. `<p/>, <table/>, <list/>, <quote/>, <ab/>`) can appear as sibling *before* a `<div/>` element but not after.


adbar:
Since we are replacing them with `ab` in #254 you could then remove the `fw` elements here as well.

adbar:
Thanks!

PR #254, opened 2022-09-27
TEI: remove text from <div> elements

knit-bee:
In TEI P5, `<div/>` elements shouldn't have text content. I added a check to find `<div/>`  elements with text. 
If such text occurs, it is either added to a new `<p/>` Element that is added as the first child of the `<div/>` or, if the first child already is a `<p/>` element, the text from the `<div/>` is joined with its text.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -648,6 +648,21 @@ def test_tei():
     assert xml.write_fullheader(header, docmeta) is not None
     docmeta.title, docmeta.sitename = None, None
     assert xml.write_fullheader(header, docmeta) is not None
+    xml_doc = etree.fromstring("<TEI><text><body><div>text</div></body></text></TEI>")
+    cleaned = xml.check_tei(xml_doc, "fake_url")
+    result = [(elem.tag, elem.text) for elem in cleaned.find(".//div").iter()]
+    expected = [("div", None), ("p", "text")]
+    assert result == expected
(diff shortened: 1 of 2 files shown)


adbar:
@knit-bee You don't appear to be testing for tails in the first element, could you please check if they are handled properly by including such a case in the tests?
`"<TEI><text><body><div><div>text1<p>text2</p>has to be there</div></div></body></text></TEI>"`

knit-bee:
I added a test case to check for a tail on `<div>` element. 

The tail on the `<p>` element would be handled by the changes in #252 . If you want, I can join the two branches together and add this example then as test case. I think the result should be 
`"<TEI><text><body><div><div><p>text1 text2 has to be there</p></div></div></body></text></TEI>"` but right now, the test would fail on both branches.

adbar:
Yes, I believe certain PRs could be merged (maybe #251, #252 and #254), could you please have a look?

knit-bee:
- I merged the changes from  #251 and #252
- I added a test for the example you mentioned before (text in `<div>` and tail on `<p>`)
- I converted all `<head>` to `<ab>` in TEI output and added a test for the rendition attribute
- I removed `<fw>` from `VALID_TEI_TAGS` in `xml.py` 

However, I was not completely sure if `<fw>` should also be removed in other places (e.g. `NEWLINE_ELEMS` in `xml.py`)

knit-bee:
I removed all instances of `<fw>` and added test cases for complex `<head>` elements.

PR #296, opened 2023-02-07
Fixed bug on JSON metadata when ld+JSON is formatted wrong

felipehertzer:
Hi @adbar,

I was testing the JSON metadata and I found this bug, the JSON is wrong, but it was causing an exception in the script. I saw your message about testing, as soon I have time I will contribute.

```json
"author":{"@type":"Person","name":{"@type":"Person","name":"John Smiths"}}
```

Article:
https://bit.ly/40KbH1w

Before:
```
AttributeError: 'dict' object has no attribute 'lower'
```

Now:
```
"author": "John Smiths"
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,24 +17,24 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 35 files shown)


adbar:
Nice catch, thanks!

PR #306, opened 2023-02-19
Add as_dict method to Document. 

### LINKED ISSUE #298: extract_metadata() doesn't return dict, but documentation says it does
Thanks for this awesome library.  

I'm using `trafilatura==1.4.1`.  

It looks like the the documentation is out of sync with the behavior of `extract_metadata()`.  

The documentation says that `extract_metadata()` returns `dict` or `None`: https://trafilatura.readthedocs.io/en/latest/corefunctions.html#extract-metadata  

But, it looks like it actually returns `trafilatura.metadata.Document`.  

I see two paths forward.  

Option 1: Update the Documentation:
- [ ] change the documentation to say that `extract_metadata()` returns `trafilatura.metadata.Document`
- [ ] (optional) add a method to `trafilatura.metadata.Document` to easily convert to a `dict`
  
Option 2: Update the Code:
- [ ] change `extract_metadata()` to return a `dict`

I'm happy to implement either option in a PR.  

Let me know what you think.

edkrueger:
* Add as_dict method to Document. 
* Add test.
* Update docstring in extract_metadata to fix docs.
* Fixes #298.

### DIFF
--- a/tests/metadata_tests.py
+++ b/tests/metadata_tests.py
@@ -275,7 +275,47 @@ def test_images():
     metadata = extract_metadata('<html><head><meta property="twitter:image" content="https://example.org/example-twitter.jpg"></html>')
     assert metadata.image == 'https://example.org/example-twitter.jpg'
 
+def test_Document_as_dict():
+    """Tests that the dict serialization works and preserves data."""
 
+    html = """
+    <html>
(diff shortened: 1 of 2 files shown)


adbar:
Thanks!

Issue #307, opened 2023-02-24
Extract page type with og:type or jd+json


andremacola:
I think it would be interesting to extract the type of page, although some sites do not use Open Graph tags.

Maybe some other algorithm as an alternative to og:type to stipulate whether it's a category/front page or articles page?

adbar:
Hi @andremacola, thanks for the suggestion, I believe it's safe for now to use `og:type`, anything else would be trickier.

If you have other ideas concerning metadata extracted from Open Graph tags feel free to write further pull requests.

andremacola:
I did a PR

adbar:
I understand, I meant other PRs if necessary.

adbar:
`article:section` and `og:locale` are also interesting if you want to look into it. I started working on it but haven't gone any further for now.

I'm closing the issue since `og:type` has been addressed. Thanks for that!

PR #310, opened 2023-02-24
feat: extract pagetype from og:type or ld+json

andremacola:
This implements https://github.com/adbar/trafilatura/issues/307

- Extract from og:type
- If no og:type tag, try to extract from ld+json
- On some sites the variable `element_text` was coming with bad html formatting causing error and having to use `extract_json_parse_error`. This patch fix the problem with `html.unescape()`
- Tests were done but I'm not sure if I should put validation in all extract_meta_json tests

OBS: I realized that json_metadata already had some `@type` implementations but I preferred not to change it to prevent break anything and implemented it with new conditions and lists (not sure if this is the ideal approach)

OBS2: I think txttocsv function and test_txttocsv should be refactored. Every minor change in metadata breaks the test.

OBS3: extract_meta_json function could be simplified and refactored

OBS4: Only returns pagetype from ld+json if the page is some kind of an article/category/home/site/bloc etc... got types from https://schema.org/docs/full.html

### DIFF
--- a/tests/metadata_tests.py
+++ b/tests/metadata_tests.py
@@ -198,7 +198,8 @@ def test_sitename():
 
 def test_meta():
     '''Test extraction out of meta-elements'''
-    metadata = extract_metadata('<html><head><meta property="og:title" content="Open Graph Title"/><meta property="og:author" content="Jenny Smith"/><meta property="og:description" content="This is an Open Graph description"/><meta property="og:site_name" content="My first site"/><meta property="og:url" content="https://example.org/test"/></head><body><a rel="license" href="https://creativecommons.org/">Creative Commons</a></body></html>')
+    metadata = extract_metadata('<html><head><meta property="og:title" content="Open Graph Title"/><meta property="og:author" content="Jenny Smith"/><meta property="og:description" content="This is an Open Graph description"/><meta property="og:site_name" content="My first site"/><meta property="og:url" content="https://example.org/test"/><meta property="og:type" content="Open Graph Type"/></head><body><a rel="license" href="https://creativecommons.org/">Creative Commons</a></body></html>')
+    assert metadata.pagetype == 'Open Graph Type'
     assert metadata.title == 'Open Graph Title'
     assert metadata.author == 'Jenny Smith'
     assert metadata.description == 'This is an Open Graph description'
--- a/trafilatura/json_metadata.py
+++ b/trafilatura/json_metadata.py
@@ -9,7 +9,7 @@ from .utils import normalize_authors, trim
 
 
 JSON_ARTICLE_SCHEMA = {"article", "backgroundnewsarticle", "blogposting", "medicalscholarlyarticle", "newsarticle", "opinionnewsarticle", "reportagenewsarticle", "scholarlyarticle", "socialmediaposting", "liveblogposting"}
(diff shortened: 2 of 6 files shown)


adbar:
Thanks for the PR, I'm going to review it. As for your remarks:

1. `@type` in metadata: @felipehertzer do you have an opinion on this?
2. Yes, `txttocsv()` could use some refactoring, I'm open for a PR on this btw.
3. I'm not sure how to refactor `extract_meta_json()`, could you provide a suggestion?
4. Sounds good.

tests/metadata_tests.py
```diff
@@ -198,7 +198,8 @@ def test_sitename():
 
 def test_meta():
     '''Test extraction out of meta-elements'''
-    metadata = extract_metadata('<html><head><meta property="og:title" content="Open Graph Title"/><meta property="og:author" content="Jenny Smith"/><meta property="og:description" content="This is an Open Graph description"/><meta property="og:site_name" content="My first site"/><meta property="og:url" content="https://example.org/test"/></head><body><a rel="license" href="https://creativecommons.org/">Creative Commons</a></body></html>')
+    metadata = extract_metadata('<html><head><meta property="og:title" content="Open Graph Title"/><meta property="og:author" content="Jenny Smith"/><meta property="og:description" content="This is an Open Graph description"/><meta property="og:site_name" content="My first site"/><meta property="og:url" content="https://example.org/test"/><meta property="og:type" content="Open Graph Type"/></head><body><a rel="license" href="https://creativecommons.org/">Creative Commons</a></body></html>')
```
adbar, line 201: Maybe provide the source as a comment?
https://schema.org/docs/full.html

trafilatura/json_metadata.py
```diff
@@ -9,11 +9,13 @@
 
 
 JSON_ARTICLE_SCHEMA = {"article", "backgroundnewsarticle", "blogposting", "medicalscholarlyarticle", "newsarticle", "opinionnewsarticle", "reportagenewsarticle", "scholarlyarticle", "socialmediaposting", "liveblogposting"}
+JSON_OGTYPE_SCHEMA = {"aboutpage", "checkoutpage", "collectionpage", "contactpage", "faqpage", "itempage", "medicalwebpage", "profilepage", "qapage", "realestatelisting", "searchresultspage", "webpage", "website", "article", "advertisercontentarticle", "newsarticle", "analysisnewsarticle", "askpublicnewsarticle", "backgroundnewsarticle", "opinionnewsarticle", "reportagenewsarticle", "reviewnewsarticle", "report", "satiricalarticle", "scholarlyarticle", "medicalscholarlyarticle", "socialmediaposting", "blogposting", "liveblogposting", "discussionforumposting", "techarticle", "blog", "jobposting"}
```
adbar, line 12: Maybe provide the source as a comment?
https://schema.org/docs/full.html

trafilatura/metadata.py
```diff
@@ -117,7 +118,7 @@ def extract_meta_json(tree, metadata):
     for elem in tree.xpath('.//script[@type="application/ld+json" or @type="application/settings+json"]'):
         if not elem.text:
             continue
-        element_text = JSON_MINIFY.sub(r'\1', elem.text)
+        element_text = html.unescape(JSON_MINIFY.sub(r'\1', elem.text))
```
adbar, line 121: How is html.unescape different from `string.encode().decode('unicode-escape')` here? We could maybe skip an import.

andremacola: `string.encode().decode('unicode-escape')` does not remove html entities like `&#13;` in my tests.

adbar: That's correct, I'm thinking of modifying the `normalize_json()` function in `json_metadata` accordingly. That way we would could use it here as well, does that work for you?

andremacola: Yep. I added a test with html entities too.

adbar: I just took care of it in 22e3075639e0d14a8ce34c87a9bff0d7d8ceab30. Please leave a message or a PR if you had something else in mind.

PR #315, opened 2023-03-13
Fix handling of text content of <div> with empty <p> child

knit-bee:
In the `check_tei` function, I fixed the bug that occurred during the cleaning of `<div/>` elements with text or tail content if there was an empty `<p/>` child present.


adbar:
Thanks!

Issue #320, opened 2023-04-09
Getting warnings while testing extract() function with with pytest


sdondley:
I have this code:

```
from trafilatura import extract
import json
from get_html_from_safari.get_html_from_safari import get_html

def get_article_data(url=None):
    if url is None:
        html = get_html()
    else:
        html = get_html(url)

    data_str = extract(html)
```

I have this test:

```
from nlp_extractor_simple.nlp_extractor_simple import get_article_data

def test_get_article():
    get_article_data();
```

When running the test, I get warnings in the pytest output:

```
tests/test_01-basic.py::test_get_article
  /usr/local/lib/python3.11/site-packages/trafilatura/core.py:560: DeprecationWarning: NotImplemented should not be used in a boolean context
    result_body.extend(filter(None.__ne__, (handle_textelem(e, potential_tags, options) for e in subelems)))

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

PR #321, opened 2023-04-09
#320 - update deprecated code

sdondley:


adbar:
Hi @sdondley, thanks for pointing that out. The lambda solution you suggest isn't by any way solving any deprecation issue though, or am I missing something?
In any case, I would rather be in favor of `filter(None, ...)`.

adbar:
I got your point and added similar changes. The lambda function may be slightly slower but keeping the code base up to date is more important. Thanks!

PR #328, opened 2023-04-21
fix: html with no metadata image

andremacola:
I noticed that the current implementation of the xpath selector may return the wrong content in some cases. Specifically, when there is no og:image property present

To fix this, I updated the xpath selector.


adbar:
Hi @andremacola, that makes perfect sense, thanks!

Issue #319, opened 2023-04-07
'lxml.etree._Element' object has no attribute 'text_content'


asjsrep:
Extraction of the following URL fails

`trafilatura -u "https://buffer.com/resources/ai-content-creation/"`

```
ERROR: 'lxml.etree._Element' object has no attribute 'text_content'
Traceback (most recent call last):
  File "venv/lib/python3.10/site-packages/trafilatura/cli_utils.py", line 339, in examine
    result = extract(htmlstring, url=url, no_fallback=args.fast,
  File "venv/lib/python3.10/site-packages/trafilatura/core.py", line 1055, in extract
    document = bare_extraction(
  File "venv/lib/python3.10/site-packages/trafilatura/core.py", line 927, in bare_extraction
    postbody, temp_text, len_text = compare_extraction(cleaned_tree_backup, tree_backup_1, url, postbody, temp_text, len_text, options)
  File "venv/lib/python3.10/site-packages/trafilatura/core.py", line 644, in compare_extraction
    algo_text = trim(temppost_algo.text_content())
AttributeError: 'lxml.etree._Element' object has no attribute 'text_content'

```

trafilatura version : 1.5.0

adbar:
Hi @asjsrep, I cannot reproduce the bug and on a conceptual level this error should not happen. So I do not know what is happening here, do you have more details to share?

asjsrep:
@adbar sure, 

I'm running it on a Mac M2, Python version Python 3.10.8

This is the pip freeze in a new virtual environment:

```
certifi==2022.12.7
charset-normalizer==3.1.0
courlan==0.9.0
dateparser==1.1.8
htmldate==1.4.2
jusText==3.0.0
langcodes==3.3.0
lxml==4.9.2
python-dateutil==2.8.2
pytz==2023.3
pytz-deprecation-shim==0.1.0.post0
regex==2023.3.23
six==1.16.0
tld==0.13
trafilatura==1.5.0
tzdata==2023.3
tzlocal==4.3
urllib3==1.26.15
```

Is there any specific debugging data which would be useful?

asjsrep:
I've had a chance to look into this in more detail now. The error was being raised in the lxml `document_fromstring` method via the `try_readability` method and appears to be due to a character encoding issue.

Changing:
`trafilatura/external.py", line 46, `

to 

` return fromstring(doc.summary().encode('UTF-8'), parser=HTML_PARSER)`

[...]

PR #331, opened 2023-04-23
Update core.py

Korben00:
When I want to use a lambda function in python like this one :

`extract_content = lambda url: trafilatura.extract(trafilatura.fetch_url(url), output_format='html', include_links=True)
`

I encounter the following error: 

`AttributeError: 'lxml.etree._Element' object has no attribute 'text_content'
`
Indeed, in the trafilatura library, the text_content() method is used to extract the text from the _Element object of lxml.etree. We can get the same result by using the etree.tostring() method with the argument method='text'.

So in the compare_extraction function of trafilatura/core.py, I replaced this line :

`algo_text = trim(temppost_algo.text_content())
`
with this line: 

`algo_text = trim(etree.tostring(temppost_algo, method='text', encoding='utf-8').decode('utf-8'))
`

Don't forget to import etree in core.py

`from lxml import etree
`
Tested and working

### DIFF
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -13,7 +13,6 @@ import re  # import regex as re
 import warnings
 
 from copy import deepcopy
-from lxml import etree
 
 # SIGALRM isn't present on Windows, detect it
 try:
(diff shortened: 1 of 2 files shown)


adbar:
Hi @Korben00, thanks, don't mind the tests above there is a problem with `httpbin.org` at the moment.

Your code looks good but I wonder if the issue isn't related to LXML's behavior on Apple M1/M2, a new version is pending, could you check if this solution fixes the bug for you: https://github.com/adbar/trafilatura/issues/319 ?

_PS: LE Monsieur Korben, enchanté ;)_

adbar:
@Korben00 I applied a variant of your fix and added a line suggested in #319, however I cannot reproduce the bug so I cannot be sure both fixes are necessary. Please get in touch if you still run into errors.

Issue #330, opened 2023-04-23
Proxy support to Trafilatura


andremacola:
Currently, I'm having trouble accessing some websites and I believe that using a proxy might help solve this issue.

If there is no natively proxy support in Trafilatura (didn't find in docs), I would like to suggest adding this functionality for future versions.

andremacola:
Apparently urllib3 has chosen not to read environment proxy variables  https://github.com/urllib3/urllib3/issues/1785

andremacola:
In https://github.com/adbar/trafilatura/blob/82043f7e84d256571cc1861249e27103193508ba/trafilatura/downloads.py#L104 we could use something like:

```python
if use_proxy:
    HTTP_POOL = urllib3.ProxyManager(retries=RETRY_STRATEGY, timeout=config.getint('DEFAULT', 'DOWNLOAD_TIMEOUT'), ca_certs=certifi.where(), num_pools=NUM_CONNECTIONS, proxy_url=PROXY_HOST, proxy_headers=PROXY_HEADERS)
else:
    HTTP_POOL = urllib3.PoolManager(retries=RETRY_STRATEGY, timeout=config.getint('DEFAULT', 'DOWNLOAD_TIMEOUT'), ca_certs=certifi.where(), num_pools=NUM_CONNECTIONS)
```

PROXY_**** variables could come from `config` or maybe from `os env variables` or params in `trafilatura.fetch_url()` if the user want to manipulate some random use of proxies.

What do you think?

adbar:
Trafilatura's download utilities should stay simple in order not to confuse users. There are lots of alternatives and downloading at scale is a different challenge altogether. A worst case solution would be to use another software for downloads and to process the resulting HTML files with Trafilatura.

That being said, if you can find a easy way to perform HTTP requests with a proxy then it could be an interesting additional feature.

fortyfourforty:
> Trafilatura's download utilities should stay simple in order not to confuse users. There are lots of alternatives and downloading at scale is a different challenge altogether. A worst case solution would be to use another software for downloads and to process the resulting HTML files with Trafilatura.

Second this. There are tons of efficient ways for downloading. Trafilatura should stay focused on its main task: extraction. 
It's not a good idea to make it more bloated with unnecessary features. 
Just use scraping tools for scraping. Trafilatura should not be a Swiss army knife tool.

Issue #210, opened 2022-05-23
trafilatura behing a proxy


hpiedcoq:
Hi,

Trafilatura doesn't seem to work behind our proxy.

Config : a docker container, with openjdk11-slim-bullseye.
A dig soa works like a charm (so internet is up).

print env says : 

```
no_proxy=*.c**************,127.0.0.0/8,10.0.0.0/8,*************12,192.168.0.0/16
HOSTNAME=***********************
JAVA_HOME=/usr/local/openjdk-11
PWD=/app
HOME=/
LANG=C.UTF-8
https_proxy=http://p********************************:xxxx
TERM=xterm
no_proxy=*.c**************,127.0.0.0/8,10.0.0.0/8,*************12,192.168.0.0/16
SHLVL=1
HTTPS_PROXY=http://p********************************:xxxx
HTTP_PROXY=http://p********************************:xxxx
http_proxy=http://p********************************:xxxx
PATH=/usr/local/openjdk-11/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/app
RELEASE=***********
DEBIAN_FRONTEND=noninteractive
JAVA_VERSION=11.0.15
```
Any idea?

adbar:
Hi @hpiedcoq, I'm afraid the question is too specific for me to reproduce or to answer.

The underlying library used for downloads is `urllib3`, I suggest you try to run a simple script with a [pool manager](https://urllib3.readthedocs.io/en/stable/advanced-usage.html) (i.e. like trafilatura) and see if it works. If it doesn't then it would be an issue related to urllib3.

Another option would be to use another download utility and then process the HTML files locally (on your docker instance) with trafilatura.

PR #332, opened 2023-04-24
feat: use proxy to extract data

andremacola:
This PR implements https://github.com/adbar/trafilatura/issues/330 and solves the issue https://github.com/adbar/trafilatura/issues/210

I have updated the downloads_tests.py file to reflect these changes, although I have not included a proxy test at this time. It maybe possible to read a private proxy variable in .env/settings.cfg for tests.

I can confirm that the HTTP/HTTPS SSL proxies are working correctly.

In order to use Socks4/5 proxies with urllib3, it is necessary to install the urllib3[socks] package using the command `python -m pip install urllib3[socks]` as described here: https://urllib3.readthedocs.io/en/stable/advanced-usage.html#socks-proxies. I did not have access to a Socks proxy for testing, I was unable to include this feature in the current pull request.

Finally, the trafilatura.fetch_url() function now includes a new parameter use_proxy, which can be used to specify a proxy in the format https://proxy_user:proxy_pass@proxy_host:proxy_port.

### DIFF
--- a/tests/downloads_tests.py
+++ b/tests/downloads_tests.py
@@ -62,13 +62,13 @@ def test_fetch():
     # doesn't work?
     # assert _send_request('https://expired.badssl.com/', False, False, DEFAULT_CONFIG) is not None
     if pycurl is not None:
-        assert _send_pycurl_request('https://expired.badssl.com/', False, DEFAULT_CONFIG) is not None
+        assert _send_pycurl_request('https://expired.badssl.com/', False, False, DEFAULT_CONFIG) is not None
     # no SSL, no decoding
     url = 'https://httpbin.org/status/200'
     response = _send_request('https://httpbin.org/status/200', True, False, DEFAULT_CONFIG)
     assert response.data == b''
     if pycurl is not None:
-        response1 = _send_pycurl_request('https://httpbin.org/status/200', True, DEFAULT_CONFIG)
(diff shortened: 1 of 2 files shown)


adbar:
Hi @andremacola, thanks but this is a substantial change, it would be better if you could find a way to test for proxy behavior. As it is you've tested it separately but you added untested lines of code. There may be a way to mock the proxy in unit tests.

andremacola:
@adbar I'll try to find some solution for tests. I admit that my knowledge in the area of proxy testing is limited.

Here are some thoughts:

1. We could create a docker container to simulate a proxy server (too much work to do)
2. Use a public proxies api just for tests (but this is not reliable)
3. Maybe find something in python to "fake a proxy"

adbar:
I agree that solution 3 would be best, the proxies don't need to be real. Solution 2 otherwise, but right now there is a problem with `httpbin.org` which I use for other tests so yes, these APIs are not always reliable.

adbar:
@andremacola Any news on the PR?

andremacola:
> @andremacola Any news on the PR?

Sorry @adbar, right now I'm using this PR in my own project but currently my main job is taking up a lot of my time. I don't have any updates regarding the tests for this.

adbar:
@andremacola Since we're not making progress right now I would close the PR, you can always re-open it later, does that work for you?

PR #352, opened 2023-05-16
simplify code for JSON metadata extraction

adbar:
@felipehertzer I tried to clean up a bit, could you please have a look?

### DIFF
--- a/trafilatura/json_metadata.py
+++ b/trafilatura/json_metadata.py
@@ -129,16 +129,17 @@ def extract_json_parse_error(elem, metadata):
     # author info
     element_text_author = JSON_AUTHOR_REMOVE.sub('', elem)
     if any(JSON_MATCH.findall(element_text_author)):
-        element_text_author = JSON_AUTHOR_REMOVE.sub('', elem)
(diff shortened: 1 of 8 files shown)


trafilatura/json_metadata.py
```diff
@@ -131,10 +129,9 @@ def extract_json_parse_error(elem, metadata):
     # author info
     element_text_author = JSON_AUTHOR_REMOVE.sub('', elem)
     if any(JSON_MATCH.findall(element_text_author)):
-        author = extract_json_author(element_text_author, JSON_AUTHOR_1)
-        if author is None:
-            author = extract_json_author(element_text_author, JSON_AUTHOR_2)
-        if author is not None:
+        element_text_author = JSON_AUTHOR_REMOVE.sub('', elem)
```
felipehertzer, line 132: Duplicated line with # 130

felipehertzer (changes requested):
Hey @adbar that's changes makes totally sense to me. Thanks.

trafilatura/json_metadata.py
adbar: Thanks!

PR #370, opened 2023-06-14
more efficient HTML parsing code

adbar:


adbar:
@felipehertzer Could you please test this and tell me if it solves the problem you described here https://github.com/adbar/trafilatura/pull/365#issuecomment-1581776131 ?

felipehertzer:
@adbar Thats is a great change, it saved 5 extra calls to fromstring_bytes only in json_metadata_tests.py

adbar:
Thanks for your feedback, I'll keep the PR as a draft to test it further before merging it.

PR #372, opened 2023-06-16
improve code block support

idoshamun:
While evaluating trafilatura, I noticed that the code blocks support in sites like StackOverflow and GitHub is lacking. 

This PR introduces support in [highlight.js](https://highlightjs.org/), the most popular framework for embedding code blocks, GitHub and W3Schools code blocks. 

I added relevant tests to evaluate this use case.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -1050,6 +1050,34 @@ def test_list_processing():
     assert target_element.tail == 'tail'
 
 
+def test_code_blocks():
+    highlightjs = '''<div class="s-prose js-post-body" itemprop="text">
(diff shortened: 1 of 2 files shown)


adbar:
Thanks, everything looks good, I'll test the PR on my benchmark and eventually merge it.

idoshamun:
@adbar thanks for the quick response!
I would like to add a real world test to validate it will work in the wild. Because after some manual testing I may have missed something.
I'll keep you posted

idoshamun:
@adbar after checking again manually everything works fine. It was just a matter of wrong config in my other project.
You can merge whenever :)

PR #377, opened 2023-06-20
Convert relative URLs in links to absolute by default

### LINKED ISSUE #376: Option to convert relative links
Would you be interested in adding an option to convert relative URLs to absolute when doing `trafilatura --links`?

feltcat:
Closes #376

If you use TXT output format, the URL isn't extracted from the text, so you must specify the URL in order for the conversion of relative URLs to work.

Let me know if you'd like any changes made.


feltcat:
So you mean just make it so that it always attempts to convert relative links, without an option to keep the current behaviour?

adbar:
Yes that looks fine. Since we're at it you could also have a look at this bug which concerns the conversion of relative links when the webpage is not the homepage:
https://github.com/adbar/courlan/issues/45

trafilatura/htmlprocessing.py
```diff
@@ -233,12 +234,17 @@ def convert_tags(tree, options):
...
+        base_url = url and get_base_url(url)
         for elem in tree.iter('a', 'ref'):
             elem.tag = 'ref'
             # replace href attribute and delete the rest
             target = elem.get('href') # defaults to None
             elem.attrib.clear()
             if target is not None:
+                # convert relative URLs
+                if base_url is not None:
+                    target = fix_relative_urls(base_url, target)
```
adbar, line 247: @feltcat please also add a test for this line

feltcat:
Added a test (sorry for the double-push!)

adbar:
Thanks, the PR now looks ready to merge.

PR #401, opened 2023-08-03
improve code support

idoshamun:
This PR improves support in detecting and handling code snippets as follows:
* pypi.org snippets support (page example: https://pypi.org/project/openai-function-call/)
* Medium scraping with and without js enabled (page example: https://medium.com/@jxnlco/seamless-integration-with-openai-and-pydantic-a-powerful-duo-for-output-parsing-fcb1e616167b)
* Keep proper formatting which includes spaces and new lines for code snippets to preserve correctness of the syntax

### DIFF
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -27,7 +27,7 @@ from .htmlprocessing import (convert_tags, handle_textnode, process_node,
                              prune_unwanted_nodes, tree_cleaning)
 from .metadata import extract_metadata, Document
 from .settings import use_config, DEFAULT_CONFIG, TAG_CATALOG
-from .utils import is_image_file, load_html, normalize_unicode, trim, txttocsv
+from .utils import is_image_file, load_html, normalize_unicode, trim, txttocsv, CODE_SPACE
 from .xml import (build_json_output, build_xml_output, build_tei_output,
                   control_xml_output, remove_empty_elements, strip_double_tags, xmltotxt)
 from .xpaths import (BODY_XPATH, COMMENTS_XPATH, COMMENTS_DISCARD_XPATH, OVERALL_DISCARD_XPATH,
@@ -225,6 +225,13 @@ def handle_code_blocks(element):
     processed_element = deepcopy(element)
     for child in element.iter('*'):
         child.tag = 'done'
+
+    # encode code spaces differently to keep formatting
+    for child in processed_element.getchildren():
+        if child.text is not None:
+            child.text = child.text.replace(' ', CODE_SPACE)
+        if child.tail is not None:
+            child.tail = child.tail.replace(' ', CODE_SPACE)
     processed_element.tag = 'code'
     return processed_element
 
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -208,7 +208,7 @@ def handle_lists(element, options):
 
 def get_code_block_element(element):
     # pip
-    if element.get('lang') is not None:
+    if element.get('lang') is not None or element.tag == 'code':
         return element
     # GitHub
     parent = element.getparent()
@@ -491,6 +491,13 @@ def recover_wild_text(tree, result_body, options, potential_tags=TAG_CATALOG):
         search_expr += '|.//div|.//lb|.//list'
     # prune
     search_tree = prune_unwanted_sections(tree, potential_tags, options)
+    # find hljs elements to detect that it's code
+    hljs_elems = search_tree.xpath("//span[starts-with(@class,'hljs')]")
+    if len(hljs_elems) > 0:
+        for e in hljs_elems:
+            parent = e.getparent().getparent()
+            if parent.tag == 'quote':
+                parent.tag = 'code'
     # decide if links are preserved
     if 'ref' not in potential_tags:
         strip_tags(search_tree, 'a', 'ref', 'span')
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -206,36 +206,30 @@ def handle_lists(element, options):
     return None
 
 
-def get_code_block_element(element):
+def is_code_block_element(element):
     # pip
     if element.get('lang') is not None or element.tag == 'code':
-        return element
+        return True
     # GitHub
     parent = element.getparent()
     if parent is not None and 'highlight' in parent.get('class', default=''):
-        return element
+        return True
     # highlightjs
     code = element.find('code')
     if code is not None and len(element.getchildren()) == 1:
-        return code
-    return None
+        return True
+    return False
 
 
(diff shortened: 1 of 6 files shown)


adbar:
Hi @idoshamun, thanks for the PR! Could you please add coverage for the line that is missing?

idoshamun:
@adbar will do!

idoshamun:
@adbar we should be ready to go!

trafilatura/utils.py
```diff
@@ -270,6 +271,8 @@ def line_processing(line):
     # remove newlines that are not related to punctuation or markup
     # remove non-printable chars and normalize space characters (including Unicode spaces)
     line = trim(remove_control_characters(LINES_TRIMMING.sub(r' ', line)))
+    # replace unique code spaces with regular space
+    line = line.replace(CODE_SPACE, ' ')
```
adbar, line 275: This function is critical for performance, is this line strictly necessary? I saw that you handle code spaces differently in `handle_code_blocks`.

trafilatura/core.py
```diff
@@ -488,6 +495,13 @@ def recover_wild_text(tree, result_body, options, potential_tags=TAG_CATALOG):
         search_expr += '|.//div|.//lb|.//list'
     # prune
     search_tree = prune_unwanted_sections(tree, potential_tags, options)
+    # find hljs elements to detect that if it's code
```
adbar, line 498: This seems more like a conversion issue and could be handled elsewhere, what do you think?

adbar:
@idoshamun Thanks for adding a corresponding test!

While your changes make perfect sense the cases tackled by them are quite specific and subject to change over time. I just want to make sure that the software can continue being used as a general-purpose extractor, the questions above arise from this fact.

idoshamun:
@adbar I'll give it a look hopefully during the weekend :)

trafilatura/utils.py
idoshamun: The problem I was trying to solve was to keep spaces within code blocks. The current version strips multiple spaces and these spaces are crucial for code (e.g indentation in python). I don't like my solution, but I wasn't able to come up with something else. Within code blocks, I replace spaces with something else and at the end return it back to spaces, this way I get around the stripping of spaces. Happy to adjust if you can recommend another way.

trafilatura/core.py
idoshamun: Yes, I agree. Can you suggest the right place, please? I'm still not highly familiar with the codebase

trafilatura/xpaths.py
```diff
@@ -144,7 +144,7 @@
     or contains(@class, "message-container") or contains(@id, "message_container")
     or contains(@class, "yin") or contains(@class, "zlylin") or
     contains(@class, "xg1") or contains(@id, "bmdh")
-    or @data-lp-replacement-content]''',
+    or @data-lp-replacement-content][not(starts-with(@class,'hljs'))]''',
```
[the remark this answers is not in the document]
idoshamun: I added it to support this line: https://github.com/adbar/trafilatura/pull/401/files/df66b25c2319af203acb43cac648803e65de3351#diff-abef8bfc97e60c84035950c7da832f5ca07580db79e4fb9e77ef0f8b82112b30R499.
I want to be able to find span elements that have a class that starts `hljs` because it means it's a code block element.
I agree that it's not ideal and see you also commented on the other comment, so maybe if we can find better place for the implementation it can make this change unnecessary

trafilatura/utils.py
adbar: I try to fix it by appending it just above.

trafilatura/core.py
adbar: I moved this code part around to use it earlier, i.e. during conversion.

trafilatura/xpaths.py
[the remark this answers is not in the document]
adbar: I need to think about it a bit more...

adbar:
@idoshamun I tried to address part of our concerns, feel free to have a look at the code.

The tests are failing right now but this is something else, I'll retry later.

idoshamun:
@adbar sorry for the delay on my side kinda busy these days. We'll do my best to review all changes asap

trafilatura/utils.py
```diff
@@ -73,6 +73,7 @@
 CLEAN_META_TAGS = re.compile(r'["\']')
 
 STRIP_EXTENSION = re.compile(r"\.[^/?#]{2,63}$")
+CODE_SPACE = ';cs;'
```
idoshamun, line 76: We don't use it anymore right?

idoshamun:
@adbar thanks for the fixes, I only have one concern. I don't see where we replace spaces with `;cs;` in code elements. I see only the other way around but the tests pass so maybe I missed something.

trafilatura/utils.py
adbar: That's right!

adbar:
It's at the end of this long line:
`line = line.replace('&#13;', '\r').replace('&#10;', '\n').replace('&nbsp;', '\u00A0').replace(';cs;', ' ')`

I still need to address the questions above but we're moving forward, thanks for your help!

idoshamun:
@adbar this long line switches the encoded space to a regular space. But I don't see where we do the other way around. It was removed in your changes but I didn't find the alternative

adbar:
@idoshamun I see what you mean, you added lines like this: `child.text = child.text.replace(' ', CODE_SPACE)`

The tests are passing though, so either it's not so important or you didn't add a corresponding test. Could you please check if the current version works for you? If not feel free to put back the corresponding lines along with a suitable test case.

idoshamun:
@adbar I think I saw in one of the commits you removed the spaces in the tests. I'll give it a look

adbar:
@idoshamun yes, my bad, it was in 52688eaec8ef9fa4a302a875380aba4147aaee9a because I didn't understand the point. Now I see.

idoshamun:
@adbar is there anything I can do at this point to help?

adbar:
@idoshamun I was able to solve most problems mentioned above and will now merge the PR as substantial progress has been made.

The spacing issue remains, it's not clear to me how to address it at this point. You could work on it in another PR, as well as add further tests if you witness issues with the robustness of code extraction.

idoshamun:
@adbar thanks a lot. Can you give me some pointers of where to look for the spaces? What will be the best place?
For some languages it's crucial to keep proper spacing

adbar:
The problem is that the spaces are stripped at some point, I can look further into it if you don't find the culprit.

idoshamun:
I'm gonna have a look now at what will be a better way to implement the spaces. Any pointers?

PR #259, opened 2022-10-19
Added the possibility to prune custom path's

HeLehm:
It is now possible to prune the tree via a list of custom defined xpath's in extract and bare_extraction.

PR #414, opened 2023-09-01
added possibility to prune xPaths

HeLehm:
Same as I did in #259 .

Also: I added a simple test.


adbar:
@HeLehm Thanks, the PR looks good, I'll hopefully merge it soon.

HeLehm:
Hi @adbar, any updates on this? Can I do something to help?

adbar:
Hi @HeLehm, I'll be afk for a little while but I'll merge your PR in October at the latest. In the meantime, could you please add the following tests? Not strictly speaking a coverage problem, rather a logical way to cover everything:

- You only test for `str` for `prune_xpath`, please also test with a list
- In your test the whole document is stripped, just add a test to make sure the right amount of text is removed

The two items can be combined.

adbar:
Thanks! I don't quite understand why the coverage changes should be an issue and I plan to merge the PR as it is.

PR #419, opened 2023-09-09
docs: fix quickstart

sashkab:
Code in quickstart imported `fetch_url` and `extract` from `trafilatura`, but then attempting to use `trafilatura.extract`, which wasn't imported.


adbar:
Hi @sashkab, good catch, thanks!

PR #425, opened 2023-10-03
Test on Python 3.12 production release

cclauss:


adbar:
Hi @cclauss, thanks for the PR, do you know if it's possible to add a development version of Python 3.13 to the tests as well?

adbar:
Can we put the PR on hold until you add `3.13-dev` or so?

cclauss:
Please approve the workflow so we can see which Python 3.12 it uses.  If it is prior to the production version (like 3.12 release candidate 3) then I would not wait.

cclauss:
Not required `3.12-dev` autoupgrades to production when it becomes available.
Python 3 12

PR #429, opened 2023-10-11
preserve space in certain elements

### LINKED ISSUE #422: Multiple spaces within a text element are not supported
Following #401, I noticed Trafilatura doesn't support multiple spaces within a text element. Specifically code element is what I care about. This causes syntax issues in some languages like python which are indent based.

I debugged the issue and realized the root cause is here: https://github.com/adbar/trafilatura/blob/master/trafilatura/utils.py#L268.
Essentially when we call trim it ignores all the multiple spaces because it split the line by spaces. I'm not sure why we need to trim multiple spaces in the first place, if it's for compression reasons than we can probably remove multiple spaces between elements and not within.

Potentially we can change this line to:
`remove_control_characters(LINES_TRIMMING.sub(r' ', line)).strip()`
But I'm not sure if it introduces new edge cases.

@adbar what would be the right way to address this issue?

idoshamun:
It's still WIP but @adbar, I need your advice and early review to see that I'm heading the right direction.
Currently I manage to preserve spaces in code elements but it breaks proper formatting in other elements. You can see the test cases. I'm not sure what will be the right way to address these cases.
It's mostly where `<p>H <hi>i</hi>, <hi>w</hi></p>`.
Basically content elements (p, item, etc) should preserve spaces as well I think

Closes #422


adbar:
Hi @idoshamun, I also think that's the idea. My remarks:
- `PRESERVE_SPACE_TAGS` would need to be a set for faster "in" operations
- Currently this functionality/line from `sanitize()` is missing from the PR:
`'\n'.join(filter(None, (line_processing(l) for l in text.splitlines())))`

idoshamun:
@adbar good catch! I applied the changes please check again

idoshamun:
@adbar would you mind suggesting the next step?

adbar:
@idoshamun Sorry, I missed your last comment. The tests don't pass, do we need to adapt them to your changes or change the code further? That would be the next step IMO.

idoshamun:
@adbar this is exactly my debate, not sure if I need to adjust the tests or my code. Can you have a look and let me know your suggestion please?
Also regardless of the tests, does the code make sense to you before I proceed further with the changes?

adbar:
Hi @idoshamun, yes, the code makes sense so I'd start with the tests, I'm not sure why they fail.

idoshamun:
@adbar it's ready for your review!

adbar:
Thanks, I need to check the impact on general performance (accuracy and speed) so it's going to take a bit of time. I'm not completely sure regarding tables but this is the way to go for code segments.

Just to clarify: you inserted `\xa0` spaces in the tests as a result of the `str.replace('&nbsp;', '\u00A0')` transformation, right?

idoshamun:
> Just to clarify: you inserted \xa0 spaces in the tests as a result of the str.replace('&nbsp;', '\u00A0') transformation, right?

Yap.

Let me know if I can help with something :)

adbar:
So, the changes pass the tests, including on my benchmark, they are also not critical for performance. Thanks for the PR! I thought we could maybe make it better:
- You're only checking the parent, it cannot be that a relevant element is more deeply nested, i.e. a tag like `<code>` is never far?
- What about `<pre>`? I don't see it in your list but would be tempted to add it to `SPACING_PROTECTED`.
- We could also let users decide by putting this list among the variables in `settings.cfg`, that's an open question, I'm thinking more and more to put constants there instead of embedded somewhere in the scripts.

idoshamun:
Thanks a lot for checking and the effort @adbar!

> You're only checking the parent, it cannot be that a relevant element is more deeply nested, i.e. a tag like <code> is never far?

I can suggest a different approach then, whenever we detect a new preserve space element, we'd propagate down an attribute to indicate we need to preserve space to every child element. This way all the tree will preserve space. Let me know if I should go down this path.

> What about <pre>? I don't see it in your list but would be tempted to add it to SPACING_PROTECTED.

You're right. I'll add the pre element to the set.

> We could also let users decide by putting this list among the variables in settings.cfg, that's an open question, I'm thinking more and more to put constants there instead of embedded somewhere in the scripts.

Once you decide to apply this strategy we can consider it. But if it's not a common pattern in the library currently than I suggest we skip this one. Wdyt?

adbar:
The way we perform tree traversal is crucial because it can save time. I was just curious if elements such as `code` are deeply nested, possibly not but I have no data on that.
Using attributes on the HTML structure might even be better but I'm not sure that preserving spaces in all the tree would be beneficial. Since your PR is already an improvement we could leave it at that and see the rest later. It also depends on the time you can invest in it.

Yes, please add `pre` to the set.

We can see the configuration/options later, no worries.

adbar:
I still get this `FutureWarning`:
`FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.`

I believe it's about statements like `(p and p.tag in SPACING_PROTECTED)`.
You could also use `if/else` statements if the lines become too long.

adbar:
Another issue, the XML output is garbled instead of pretty-printed because of the PR but there is not test for that and I'm not sure why.

adbar:
OK, I'll take it from here and try to restore parts of the removed code and to avoid the warnings.

idoshamun:
Thanks a lot @adbar for the support!
When is the next release?

PR #436, opened 2023-10-31
add_raw_html_prop

HawkClaws:
The main content extraction capability of "trafilatura" is great, 
but it is a waste that the final output is text. 
The raw_html property has been modified to return the HTML after the main content has been extracted.

### DIFF
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -15,6 +15,7 @@ from copy import deepcopy
 
 from lxml.etree import Element, SubElement, strip_elements, strip_tags
 from lxml.html import tostring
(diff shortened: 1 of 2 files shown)


adbar:
Hi @HawkClaws, thanks but unfortunately the change you implemented is a duplicate of the `output_format="xml"` option as the HTML is converted to XML during processing. So the PR doesn't appear to be useful at this point.

HawkClaws:
Oops, you were already able to output with output_format="xml",
I should have read the documentation properly... shame on me!
Thank you for pointing that out.
I'll close it.

PR #439, opened 2023-11-06
Update tutorial-epsilla.rst

richard-epsilla:
Change the performance differentiation to be accurate


adbar:
Hi @richard-epsilla, I'm currently updating the docs and I've committed your change to that branch, I will merge it shortly.

Issue #450, opened 2023-12-01
Cannot fetch url with more than 2 redirections


julienlambert42:
Hello,

The fetch_url() function returns None when a redirection url is used as argument.

For example : 
[Redirecting url](https://www.hydrogeninsight.com/production/breaking-us-reveals-the-seven-regional-hydrogen-hubs-to-receive-7bn-of-government-funding/2-1-1534596) returns None because this url redirects to [Actual url](https://www.hydrogeninsight.com/production/us-reveals-the-seven-regional-hydrogen-hubs-to-receive-7bn-of-government-funding/2-1-1534596)

The requests.get(url).text function from requests library allow redirections.

--> Is this behaviour done on purpose (for security reasons for example) ? 
--> If yes, could you add an argument to the function to allow redirections (actually it would even be better to have a num_redirection argument set to 0 by default, so we can set a threshold to avoid scam redirections)

Thanks !

Issue #128, opened 2021-10-29
trafilaturan.fetch_url failed to terminate


Guest400123064:
Hi,

I am a graduate student developing a data pipeline for semantic analysis over news articles and I just encountered this great package. However, after several experiments, there seem to be some problem with `trafilaturan.fetch_url` with the following [link](http://naplesherald.com/2018/02/07/thousands-african-migrants-protest-israel-deportations/) or `"http://naplesherald.com/2018/02/07/thousands-african-migrants-protest-israel-deportations/"`

After digging into the source code in the `downloads.py`, I discovered that `HTTP_POOL.request('GET', url, headers=_determine_headers(config))` under function `_send_request()` cannot terminate. I wonder if it is caused by the `RETRY_STRATEGY` setting because if I fetch the above URL using `response = requests.get`, the `response.status_code` is `503`. My current workaround is simply to fetch the HTML myself. I wonder if there is any way that I can set a timeout?

```
python version = '3.9.7'
trafilatura.__version__ = '0.9.3'
urllib3.__version__ = '1.26.7'
```

Thank you!

Best,
Yuxuan

vbarbaresi:
Hi,
I was able to reproduce the issue
The bug happens with the `http://` URL, which actually returns `301` response.
Note that `requests.get` automatically follows redirects (unless you set `allow_redirects=False`), so it can mask some redirection issues.

```
trafilatura -u http://naplesherald.com/2018/02/07/thousands-african-migrants-protest-israel-deportations/ -vv
```
It's stuck after this line:
```
DEBUG:urllib3.util.retry:Incremented Retry for (url='/2018/02/07/thousands-african-migrants-protest-israel-deportations/'): Retry(total=1, connect=None, read=None, redirect=None, status=None)
```
If we use the `https://` URL, it stops immediately after the `503` response.

```
trafilatura -u https://naplesherald.com/2018/02/07/thousands-african-migrants-protest-israel-deportations/ -vv

ERROR:trafilatura.downloads:retries/redirects: https://naplesherald.com/2018/02/07/thousands-african-migrants-protest-israel-deportations/ HTTPSConnectionPool(host='naplesherald.com', port=443): Max retries exceeded with url: /2018/02/07/thousands-african-migrants-protest-israel-deportations/ (Caused by ResponseError('too many 503 error responses'))
[...]

PR #461, opened 2023-12-30
introduce `MAX_REDIRECTS` config setting and fix urllib3 redirect handling

vbarbaresi:
Fixes issue #450

After setting `MAX_REDIRECTS` to 5, I could fetch the original URL from the issue: `trafilatura -u https://www.hydrogeninsight.com/production/breaking-us-reveals-the-seven-regional-hydrogen-hubs-to-receive-7bn-of-government-funding/2-1-1534596`

I also fixed this old issue: https://github.com/adbar/trafilatura/issues/128
The underlying urllib3 bug has not been fixed: `github.com/urllib3/urllib3/issues/2475`

I had to pass the retry strategy to the actual request method: it doesn't propagate from the pool maanger


trafilatura/downloads.py
```diff
@@ -90,8 +89,8 @@ def _send_request(url, no_ssl, config):
     global HTTP_POOL, NO_CERT_POOL, RETRY_STRATEGY
     if not RETRY_STRATEGY:
         RETRY_STRATEGY = urllib3.util.Retry(
-            total=0,
-            redirect=MAX_REDIRECTS, # raise_on_redirect=False,
+            total=config.getint("DEFAULT", "MAX_REDIRECTS"),
```
vbarbaresi, line 92: I had to allow at least as many retries as the max redirects here.
I wonder if we should set max_redirects to 0 as the new default because it was effectively 0 as it didn't work well

vbarbaresi:
Tests are failing with the following:

```
   def test_ab_with_p_parent_resolved():
        parser = XMLParser(remove_blank_text=True)
        xml_doc = fromstring("<text><p><head>text1</head></p></text>")
        cleaned = check_tei(xml_doc, "fake_url")
        assert cleaned.find(".//ab") is not None and cleaned.find(".//p") is None
        xml_doc = fromstring("<body><p>text1<head>text2</head></p></body>")
        cleaned = check_tei(xml_doc, "fake_url")
        result = cleaned.find(".//ab")
        assert result.getparent().tag == "body" and result.text == "text2"
        xml_doc = fromstring("<TEI><text><body><p><head>text1</head></p>text2</body></text></TEI>")
        cleaned = check_tei(xml_doc, "fake_url")
>       assert cleaned.find(".//ab").text == "text1" and cleaned.find(".//p").text == "text2"
E       AttributeError: 'NoneType' object has no attribute 'text'

tests/xml_tei_tests.py:285: AttributeError
```

I'm not reproducing locally and I don't get how my changes could have affected this part 🤔

vbarbaresi:
oh no.. lxml 5.0 just got released yesterday: https://pypi.org/project/lxml/#history
I'll pin lxml to < 5.0 for now.

adbar:
It works, thanks, but does it really fix the issue https://github.com/urllib3/urllib3/issues/2475 ? Merging this PR would close it.

vbarbaresi:
> does it really fix the issue https://github.com/urllib3/urllib3/issues/2475 ? Merging this PR would close it.

It doesn't, but will it close it? I wouldn't think merging a PR could close an issue living in another repository

adbar:
You could unlink it to be sure (I cannot).

vbarbaresi:
done, I edited the description to quote the issue URL

trafilatura/downloads.py
adbar: It's better to keep the intended logic.

trafilatura/downloads.py
```diff
@@ -327,3 +326,12 @@ def _send_pycurl_request(url, no_ssl, config):
     # tidy up
     curl.close()
     return RawResponse(bufferbytes, respcode, effective_url)
+
+
+def _reset_global_objects():
```
adbar, line 331: Is there a way to include the function directly in the tests instead of here?

vbarbaresi: done in 178a44195af480294d62e4d3a6a3766cd9dc0929
Interestingly, mutating a global from another module is not so trivial. If you import it using `from module import X`: it doesn't work as it becomes part of the local module scope.
I had to use
```python
trafilatura.downloads.HTTP_POOL, trafilatura.downloads.NO_CERT_POOL, trafilatura.downloads.RETRY_STRATEGY = None, None, None
```

PR #484, opened 2024-01-23
Handle changed behaviour of `lxml` `addnext` method

knit-bee:
I removed the two usages of `addnext` and replaced them by `insert`. The code is maybe a bit harder to read now, but I thought this would be easier than to check the `lxml` version or how tails are handled by the `addnext` method. 

I had to adjust two test cases, since some elements now have whitespace instead of `None` as tail. However, I thought this to be an acceptable change in behaviour.

I tested the changes with `lxml==4.9.4` and `lxml==5.1.0` and it works with both versions.


adbar:
Everything works great, thanks!

Issue #4, opened 2020-01-09
Test trafilatura on further web pages and report bugs


adbar:
I have mostly tested `trafilatura` on a set of English, German and French web pages I had run into by surfing or during web crawls. There are definitely further web pages and cases in other languages for which the extraction doesn't work so far.

Corresponding bug reports can either be filed as a list in an issue like this one or in the code as XPath expressions in [xpaths.py](https://github.com/adbar/trafilatura/blob/master/trafilatura/xpaths.py) (see `BODY_XPATH` and `COMMENTS_XPATH` lists).

Thanks!

adbar:
Closing for now.

adbar:
Extraction bugs in text and metadata can be listed here as in https://github.com/adbar/htmldate/issues/8 where issues specifically related to dates should be documented.

cheezman34:
Words are getting smashed together on this page:

https://research.checkpoint.com/2021/a-deep-dive-into-doublefeature-equation-groups-post-exploitation-dashboard/

![Screen Shot 2022-02-11 at 11 42 46 AM](https://user-images.githubusercontent.com/3431603/153658925-04af1491-325b-442b-85eb-f31bfa0e7590.png)
![Screen Shot 2022-02-11 at 11 44 11 AM](https://user-images.githubusercontent.com/3431603/153659038-76b7fcaa-27e6-4d2a-970e-d9d63f6cf962.png)

I looked into the extraction code here a bit. The date here is inside a span, which gets stripped, and then the date becomes the tail of the header. All of the whitespace (which includes a newline) gets lost, and then the tail is just directly appended to the header. I'm not sure if the best strategy to fix would be to include a space between the tail and text of the header node when they get extracted, or maybe to look for newlines in the text and somehow respect them. It looks like some of the lxml stuff just strips whitespace automatically when you access "text" and "tail" attributes.

![Screen Shot 2022-02-11 at 11 44 51 AM](https://user-images.githubusercontent.com/3431603/153659121-7d028a85-8c13-418d-acfb-4db18c951deb.png)
![Screen Shot 2022-02-11 at 11 45 15 AM](https://user-images.githubusercontent.com/3431603/153659195-e72a2547-8cdc-4f9c-9f7f-6f965491eed4.png)

I didn't dig into this one, but I'm guessing it's something similar as the first case. The webpage relies on whitespace that gets stripped by the extraction algorithm.

adbar:
[...]

PR #487, opened 2024-01-24
Merge multiple nodes returned by XPath

hugoobauer:
A pull request following our exchange:  https://github.com/adbar/trafilatura/issues/4#issuecomment-1898530601
The goal is to extract the content scattered in several <article> nodes.

Before :
trafilatura recall
`{'true positives': 2035, 'false positives': 231, 'true negatives': 2019, 'false negatives': 201, 'time': 10.339057922363281}
`
After :
trafilatura recall
`{'true positives': 2032, 'false positives': 253, 'true negatives': 1997, 'false negatives': 204, 'time': 10.399115324020386}`

In my search for differences, I noticed that there were potential problems with the document tests.
For example, in the document `bmjv.de.konsum.html`, there is "Transparenz bei Preisanpassungen" in the "without" checks, which is found in several spots in the HTML, including in the correctly extracted content. It therefore mistakenly triggers a false positive. I wanted to make this feedback on it before continuing to check the differences in the tests due to the committed changes. 
Capture d’écran 2024-01-24 à 11 24 55
Capture d’écran 2024-01-24 à 11 24 37
Capture d’écran 2024-01-24 à 11 24 30

I also disabled readabilipy in tests because I couldn't get it to work, and fixed a deserialization problem in `run_resiliparse`.

### DIFF
--- a/tests/comparison.py
+++ b/tests/comparison.py
@@ -29,7 +29,7 @@ from inscriptis import get_text
 # from libextract.api import extract as lib_extract
 from newspaper import fulltext
 from newsplease import NewsPlease
-from readabilipy import simple_json_from_html_string
+# from readabilipy import simple_json_from_html_string
 from readability import Document
 from resiliparse.extract.html2text import extract_plain_text
 from resiliparse.parse.encoding import bytes_to_str, detect_encoding
@@ -290,7 +290,7 @@ def run_boilerpipe(htmlstring):
     try:
         content = boilerpipe_extractor.get_content(htmlstring)
         # sanitize(boilerpipe_extractor.get_content(htmlstring))
-    except Exception:
+    except Exception as e:
         #print('Boilerpipe exception:', err)
         content = ''
     return content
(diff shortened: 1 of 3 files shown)


adbar:
Hi @hugoobauer, thanks for the PR and your feedback.

You're welcome to try out the evaluation script and make changes. The same goes for the data, your example is telling, at least in the case (and any others you can find) the text segment should be replaced by another.

If I understand your results correctly your modification degrades accuracy by adding false positives? Even in `favor_recall` mode?

hugoobauer:
Yes, the change adds false positives because some web pages have multiple <article> nodes, used for related content like article recommendations.
The changes only have an impact when favor_recall=True due to the added condition.

adbar:
Then we probably need to think twice about the change because it doesn't bring much on the evaluation dataset. Maybe another heuristic is needed, or we could expose your changes to the end user through an additional parameter.

It would be nice if you'd separate the changes to the evaluation from the ongoing work. If you open another pull request focusing only on the evaluation I can merge it sooner and we can continue looking for a solution.

hugoobauer:
Sure I can create 2 separated branches.
What about readabilipy that I disabled?

For the original problem, it's tricky because there are more websites that are well implemented than websites with "broken" architecture. I've done a few manual checks, and it fixes a few extractions, but in most cases it takes article suggestions.
Some heuristics ideas :
- minimum length check for each node
- an algorithm based on different criteria like link density / text length. Small texts leading to other articles would therefore be excluded.
What do you think ?

I've just tested the first idea by adding a minimum length for each node to avoid retrieving article suggestions, and the score is better.

```
Before any changes:
{'true positives': 2035, 'false positives': 231, 'true negatives': 2019, 'false negatives': 201, 'time': 11.863584518432617}

First suggestion :
{'true positives': 2032, 'false positives': 253, 'true negatives': 1997, 'false negatives': 204, 'time': 10.399115324020386}

-> Now:
{'true positives': 2011, 'false positives': 232, 'true negatives': 2018, 'false negatives': 225, 'time': 11.904361009597778}
```

The "true positives" is lower, probably because now it skips nodes with a length < 250 (MIN_EXTRACTED_SIZE), so it misses some texts. But now, false positives and true negatives are closer to the initial value.

I'll be on leave starting tomorrow for a week, I'll pick this up when I get back ;)

adbar:
- Yes, just create another branch, you can comment out the readabilipy function. I'll test it and see if I can make it work, if not we'll remove it.
- The length check is a good idea, the rationale would be "if there is much more text to be found in further `<article>` tags than in the first one, take all `<article>` tags, if not (teasers) only take the first one". 
- The tests are now failing but it's not a problem at this stage, feel free to work on it later on, no hurry.

adbar:
@hugoobauer Are you still working on it or should I close the PR for now?

hugoobauer:
Hi @adbar, sorry I'm too busy with work right now, I don't have time to work on it. You can close it for now, hopefully I'll be able to work on it later.

PR #498, opened 2024-02-05
add code formatting in TXT/Markdown output

adbar:


igiloh-pinecone:
Just tried this code - it doesn't distinct between inline `code formatting` and code blocks, which causes it to split these inline code formatted words to separate paragraphs.

adbar:
@igiloh-pinecone Thanks for the hint, I'm going to work on the problem.

adbar:
@igiloh-pinecone Does the PR work for you now?

Issue #489, opened 2024-01-26
Make markdown an explicit output format


adbar:
So far markdown is only activated when formatting is preserved. However it makes sense to improve markdown output and make it an explicit output format. TXT output would be left as it is now.

Related to #467.

mike-clark-8192:
I agree. Markdown has become so important these days.  Making it its own explicit format would allow grouping Markdown-specific configuration options together a bit easier in documentation and type hints, and allow for growth in that area.  It's a bit confusing that some output formatting options apply to plain text, some apply to Markdown, and some apply to both. Which makes sense conceptually, but it would be nice if it was sectioned out more explicitly.

PR #499, opened 2024-02-05
hack: add symbol to preserve vertical spacing

adbar:


adbar:
Adding a unicode hack is a disruptive change I think it might be better to apply it to markdown output only (#489).

PR #507, opened 2024-02-12
feeds: also use feedparser if available

adbar:


adbar:
I cannot see any change on my feed list so I'll put the PR aside for now.

PR #521, opened 2024-03-12
perf: avoid to load metadata causing perf leak

Spasfonx:


Spasfonx:
I didn't want to push this on the main repository, is a hack

PR #528, opened 2024-03-21
fix formatting by correcting order of element generation, space handling

### LINKED ISSUE #404: Corrupted Markdown output when TXT+formatting
I wrote a fairly complicated testcase.. then realized I could use the command line tool :-D

The docs indicate Markdown is an option
https://github.com/adbar/trafilatura/blob/d78fbb5e0d88566cb1326f04210a93b46db8ac87/docs/usage-python.rst?plain=1#L71

 * The plain text output (no Markdown) looks good.
 * In the examples I've tried so far the Markdown output is not usable, it appears to have the same content as text BUT the formatting is incorrect, new paragraph (line) breaks appear at odd places (e.g. the 2nd character on a line).

# Demo

## Session 1 - server test data

Get test data (once) and serve it to avoid repeatedly hitting web site (I could not see a way to pass in a file to trafilatura)

    wget -O wget_output.html http://www.pcgamer.com/2012/08/09/an-illusionist-in-skyrim-part-1/
    echo http://localhost:1234/wget_output.html
    python3 -m http.server 1234

## Session 2 - scrape data

    cd /tmp
    mkdir trafilatura_demo
    cd trafilatura_demo/

    python3 -m venv py3venv
    . py3venv/bin/activate
    python -m pip install trafilatura

    trafilatura --version

Then:

    # good text output, without formatting
    trafilatura -u http://localhost:1234/wget_output.html 

    # not great - some new lines show up
    trafilatura --links -u http://localhost:1234/wget_output.html 
    trafilatura --links --images -u http://localhost:1234/wget_output.html 

    # messed up parapgraphs and newlines in markdown
    trafilatura --formatting --links --images -u http://localhost:1234/wget_output.html 
    trafilatura --formatting -u http://localhost:1234/wget_output.html 

Partial extract showing problem:

    In
    [Skyrim]...
    ....
    "
    *Legends ....

There are others in the same document but I'm reluctant to include too much of the content. Hopefully the test case above is enough to reproduce for other people.

It's really obvious there is odd formatting when converting back into html (e.g. using pandoc in gfm mode, or any other md2html tool).

------

There is no option for html (only xml) which was my idea for a workaround.

I did poke around the code but I can;t get a handle on why white space is being injected into the xml cleaning code (I can see there are reasons for it, my ham fisted attempt to remove them all was unsuccessful :-D).

Thanks for making this tool available, I'm using the python readability module and trafilatura does a much better job at the meta data extraction (so far, readability works better for me for content extraction). I'm not sure if I'm misusing the the library.

dlwh:
Fixes #404 and a bunch of other associated issues around newlines and ordering output when formatting=true. I suspect this fixes most of the issues with formatting==True


trafilatura/xml.py
```diff
@@ -269,34 +272,50 @@ def merge_with_parent(element, include_formatting=False):
...
+    # this is text that comes after the closing tag, so it should be after any NEWLINE_ELEMS
+    if element.tail is not None:
+        returnlist.append(element.tail)
+
 def xmltotxt(xmloutput, include_formatting):
     '''Convert to plain text format and optionally preserve formatting as markdown.'''
     returnlist = []
-    # strip_tags(xmloutput, 'div', 'main', 'span')
-    # iterate and convert to list of strings
-    for element in xmloutput.iter('*'):
```
dlwh, line 277: mostly we just made this recursive so that text is processed in the right order

adbar:
Hi @dlwh, thanks for the PR, it seems really helpful!

I see a small regression on my benchmark (see readme in `tests/`) in terms of accuracy and speed, the latter is probably related to your change in the trimming function, it is necessary?

As for the accuracy I'll try to see if I find something. Integrating the PR bit by bit (say function by function) could also be useful as the changes could reach further than expected.

dlwh:
oh sorry the trim part can be reverted, missed that when cleaning up!

The spaces change is necessary because the spaces trimming is incorrect when outputting with formatting. Two words can get merged into one, etc.

adbar:
I just tried the `trim()` part and it slows things a bit but that's not all, we need to check what's happening function-by-function.

I can have a look at it next week, feel free to work on the PR if you wish to add or amend anything.

dlwh:
i reverted the changes to trim. My bad!

dlwh:
something seems wrong on master? (I ran `pip install -r eval-requirements.txt`)

```
Your branch is up to date with 'origin/master'.
~/src/trafilatura/tests (master)$ python comparison.py
(node:42555) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Traceback (most recent call last):
  File "/Users/dlwh/src/trafilatura/tests/comparison.py", line 613, in <module>
    result = run_resiliparse(htmlstring)
  File "/Users/dlwh/src/trafilatura/tests/comparison.py", line 346, in run_resiliparse
    decoded = bytes_to_str(htmlstring, detect_encoding(htmlstring))
TypeError: Argument 'data' has incorrect type (expected bytes, got str)
```

adbar:
Thanks for the change!

I don't run the evaluation script on a regular basis so it could be that lines need to be changed to account for updated dependencies or functions.
For the time being you can run a minimal version of it to focus on Trafilatura only. If I compare the master version of Trafilatura with this PR I see a slight drop in the results. Nothing worrying as such but I cannot explain it which and I'd like to pinpoint the change which leads to it.

trafilatura/utils.py
```diff
@@ -313,6 +313,9 @@ def sanitize_tree(tree):
 @lru_cache(maxsize=1024)
 def trim(string):
     '''Remove unnecessary spaces within a text string'''
+    if string is None:
```
adbar, line 316: This is superfluous considering the try/except AttributeError below

trafilatura/xml.py
```diff
@@ -211,7 +211,8 @@ def validate_tei(xmldoc):  # , filename=""
 
 
 def replace_element_text(element, include_formatting):
-    '''Determine element text based on text and tail'''
+    '''Determine element text based on **just the text** of the element. You must deal with the tail separately.'''
+    elem_text = element.text
```
adbar, line 215: Is there a good reason to define an additional variable instead of `element.text`?

trafilatura/xml.py
```diff
@@ -269,34 +272,50 @@ def merge_with_parent(element, include_formatting=False):
...
 
+def process_element(element, returnlist, include_formatting):
+    # Process children recursively
+    if element.text is not None:
+        # this is the text that comes before the first child
+        textelement = replace_element_text(element, include_formatting)
+        returnlist.append(textelement)
+
+    for child in element:
+        if child is not None:
```
adbar, line 283: By definition, `list(element)` in LXML should return existing elements, is there a problem if we remove this line?

adbar:
Hi @dlwh, I checked the code again, could you please address the questions above?

Also, I still don't quite understand the changes made to the `xmltotxt()` function, maybe the following example is too simple to grasp the point you're addressing but it seems to me that `doc.iter("*")` already returns all elements in the right order.

```
from lxml import etree

doc = etree.fromstring("<doc><a/><b><c>Test</c></b><d/></doc>")
assert list(doc) == doc.getchildren()  # first level, neither root nor children: a, b, d
assert doc.xpath("//*") == list(doc.iter("*"))  # depth-first, all elements: doc, a, b, c, d
```

Since the tests pass and your PR handles more complex elements (e.g. formatting) I'll integrate the PR once the open questions have been addressed, I'd just like to understand the interest of the recursive function.

What do you think?

trafilatura/utils.py
dlwh: yes i added it partially for the debugger and partially to be explicit. can remove though

dlwh:
the issue is with tail and "post formatting". The current code does a preorder traversal of the tree, handling .text and .tail at the same time. The correct order for is `pre_format + e.text + ''.join(recurse(c) for c in e.children) + post_format + e.tail` but the current code does roughly `pre_format + e.text + e.tail + post_format + ''.join(recurse(c) for c in e.children)`

The code happens to be correct for no-formatting because the tree is highly flattened, but once you preserve formatting it behaves badly.



As an example:

```python
def test_recurse_formatting():
    htmlstring = '''
    <html><body><article>
    <p> this is a <b>bold </b> test of the <i>italic</i> function</p>
    <a href="test.html">this is a link</a>
    </article></body></html>
    '''

    result = extract(htmlstring, no_fallback=True, output_format='txt', config=ZERO_CONFIG, include_formatting=True, include_links=True)
    print(result)
```

In master, this produces 

```
this is a
**bold ** test of the *italic* function [this is a link](test.html)
```

while in my patch, this produces

```
this is a **bold ** test of the *italic* function
[this is a link](test.html)
```

This still isn't quite right (there should be another newline) but it's better.

All that said, I've decided to move to readability-->html2text. I think trafilatura is overall marginally better at extracting content, but readability does a much better job at preserving formatting (even comparing to Trafilatura's XML component)

I do appreciate the work you've done. I think this is a good project, it just doesn't quite have the functionality I need.

I think misprocessing of `tail` is a pretty pervasive problem in this code base, fwiw. For instance, in main (and in this patch), this text gets reordered at the XML level:


```
def test_recurse_formatting():
    htmlstring = '''
    <html><body><article>
    <p> this is a <a href="test.html"><b>bold </b>zz</a> test of the <i>italic</i> function</p>
    <a href="test.html">this is a link</a>
    </article></body></html>
    '''

    result = extract(htmlstring, no_fallback=True, output_format='xml', config=ZERO_CONFIG, include_formatting=True, include_links=True)
    print(result)
```

this produces

```
<doc categories="" tags="" fingerprint="9af9dbb52a4f0283">
  <main><p> this is a <ref target="test.html"> test of the bold zz</ref></p><hi rend="#i">italic</hi>function<p><ref target="test.html">this is a link</ref></p></main>
  <comments/>
</doc>
```

Note that "test of the" got moved to the beginning of the ref somehow, while it should be inside in the ref's tail.

adbar:
Thanks for the detailed explanation and thanks again for the PR.

There is indeed a problem with nested elements in certain cases, this comes from the fact that I don't primarily focus on formatting or links, these features were added on request. The tests cover most difficulties but not all. The issue with element tails is also deeply linked to LXML, a package which is mainly beneficial but is also difficult to work with at times.

As you say, extraction accuracy depends on one's particular use case, software packages merely try to strike a balance. Existing evaluations focus on html2txt conversion which entails some structure (e.g. paragraph text not garbled, list elements correctly separated, etc.) though not in a full-fledged XML way.
Readability works OKish for English but [external evaluations](https://trafilatura.readthedocs.io/en/latest/evaluation.html) seem to agree on the fact that Trafilatura does a better job at extracting text. I recently found [this evaluation](https://github.com/dobbersc/fundus-evaluation) on English news sources and Trafilatura compares favorably to the rest, even before work in PR #530.

In any case, feel free to follow the development of Trafilatura as issues with formatting could get fixed over time.

Issue #531, opened 2024-03-29
List element inside a table is lost


mikhainin:
<details>
<summary>source.html</summary>
```html
<!DOCTYPE html>
<html lang="en-US" prefix="og: http://ogp.me/ns#"
      style="margin: 0;padding: 0; border: 0; font-size: 100%; font: inherit; vertical-align: baseline;">
<head>

</head>
<body class=""
      style="margin: 0; padding: 0; border: 0; font-size: 100%; font: inherit; vertical-align: baseline; line-height: 1;">

        <div class="biz-details-page-container-outer__09f24__pZBzx css-1qn0b6x">
            <div class="biz-details-page-container-inner__09f24__L9S07 css-1qn0b6x">
                <section class="css-1wrui0y"></section>
                <section class="css-1wrui0y"></section>
                <section class="css-afumm9"></section>
                <div class="css-s97lou">
                    <div class="css-kmgt1v" data-testid="main-content">
                        <main class=" css-1xykegj" id="main-content">

                            <div class=" css-1qn0b6x" id="location-and-hours">
                                <section class=" css-ufd2i" aria-label="Location &amp; Hours">

                                    <div class="arrange__09f24__LDfbs gutter-4__09f24__dajdg css-1qn0b6x">

                                        <div class="arrange-unit__09f24__rqHTg arrange-unit-fill__09f24__CUubG css-1qn0b6x">
                                            <div class=" css-1qn0b6x">
                                                <div class=" css-1qn0b6x">
                                                    <table class="hours-table__09f24__KR8wh css-n604h6">
                                                        <tbody class="">
                                                        <tr class="hours-table-row-space__09f24__chJx9 css-29kerx"></tr>
                                                        <tr class=" css-29kerx">
                                                            <th class="" scope="col"><p
                                                                    class="day-of-the-week__09f24__JJea_ css-ux5mu6"
                                                                    data-font-weight="bold">Mon</p></th>
                                                            <td class="css-1hgawz4">
                                                                <ul class=" list__09f24__ynIEd">
                                                                    <li class=" css-1qn0b6x"><p
                                                                            class="no-wrap__09f24__c3plq css-1p9ibgf"
                                                                            data-font-weight="semibold">10:00 AM - 8:00
                                                                        PM</p></li>
                                                                </ul>
                                                            </td>
                                                            <td class="css-1hgawz4"><span
                                                                    class="open-status__09f24__YH9PK no-wrap__09f24__c3plq css-syqjjh"
[...]

PR #534, opened 2024-04-02
Fixed lists inside tables when include_tables=True

mikhainin:
A possible fix for #531

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -1028,8 +1028,7 @@ def test_table_processing():
         (el.tag, el.text) if el.text is not None and el.text.strip() else el.tag
         for el in processed_table.iter()
     ]
-    # assert result == ["table", "row", "cell", ("p", "a list"), "list", ("item", "one"), ("item", "two"),]
-    assert result == ['table', 'row', 'cell', ('p', 'a list'), 'list']
+    assert result == ["table", "row", "cell", ("p", "a list"), 'list', ("item", "one"), ("item", "two"),]
     broken_table = html.fromstring("<table><td>cell1</td><tr><td>cell2</td></tr></table>")
     processed_table = handle_table(broken_table, TAG_CATALOG, options)
     result = [el.tag for el in processed_table.iter()]
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -1023,12 +1023,21 @@ def test_table_processing():
     </td>
     </tr></table>
     """)
-    processed_table = handle_table(table_with_list, TAG_CATALOG, options)
+    processed_table = handle_table(copy(table_with_list), TAG_CATALOG, options)
+    result = [
+        (el.tag, el.text) if el.text is not None and el.text.strip() else el.tag
+        for el in processed_table.iter()
+    ]
+    assert result == ['table', 'row', 'cell', ('p', 'a list'), 'list']
+
+    options.recall = True
+    processed_table = handle_table(copy(table_with_list), TAG_CATALOG, options)
     result = [
         (el.tag, el.text) if el.text is not None and el.text.strip() else el.tag
         for el in processed_table.iter()
     ]
     assert result == ["table", "row", "cell", ("p", "a list"), 'list', ("item", "one"), ("item", "two"),]
+
     broken_table = html.fromstring("<table><td>cell1</td><tr><td>cell2</td></tr></table>")
     processed_table = handle_table(broken_table, TAG_CATALOG, options)
     result = [el.tag for el in processed_table.iter()]
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -391,7 +391,7 @@ def handle_table(table_elem, potential_tags, options):
                             child.tag = 'cell'
                         processed_subchild = handle_textnode(child, options, preserve_spaces=True, comments_fix=True)
                     # todo: lists in table cells
-                    elif child.tag == "list":
+                    elif child.tag == "list" and options.recall:
                         processed_subchild = handle_lists(child, options)
                         if processed_subchild is not None:
                             newchildelem.append(processed_subchild)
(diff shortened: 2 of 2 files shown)


adbar:
@mikhainin It turns out your fix doesn't work well for nested elements in tables (see tests). Could you please have a look at it and see if you find a solution?

mikhainin:
@adbar, From what I can see, this test-case is failing:
```python
htmlstring = html.fromstring(
        """<html>
              <body><article>
                <table>
                  <tbody>
                    <tr>
                      <td>
                        <small>text<br></small>
                        <h4>more_text</h4>
                      </td>
                      <td><a href='link'>linktext</a></td>
                    </tr>
                  </tbody>
                </table>
              </article></body>
            </html>"""
    )
    processed = extract(
        htmlstring, no_fallback=True, output_format='xml', config=DEFAULT_CONFIG, include_links=True
    )
    result = processed.replace('\n', '').replace(' ', '')
    #assert """<table><row><cell>text<head>more_text</head></cell></row></table>""" in result
```

New version produces:
```xml
<doccategories=""tags=""fingerprint="576b5da16c181e08"><main>
<table>
    <row>
        <cell>text
            <head>more_text</head>
        </cell>
        <cell>
            <p>linktext</p>
        </cell>
    </row>
</table>
</main><comments/></doc>
```
The current **master** (expected behaviour I suppose):
```xml
<doccategories=""tags=""fingerprint="d7ffdfa76c785e1c"><main>
<table>
    <row>
        <cell>text
            <head>more_text</head>
        </cell>
    </row>
</table>
</main><comments/></doc>
```

Could you explain why `<p>linktext</p>` should not exist in the output?

adbar:
Yes, you're perfectly right, we need to change this test then. Do you want to do it? By the way, you could also add a test for the particular problem you're solving.

Your PR slightly decreases precision on [the benchmark](https://github.com/adbar/trafilatura/blob/master/tests/README.rst), it could have something to do with undesirable content getting added by `.itertext()`, maybe we need to filter the nested elements more.

mikhainin:
> Your PR slightly decreases precision on [the benchmark](https://github.com/adbar/trafilatura/blob/master/tests/README.rst), it could have something to do with undesirable content getting added by .itertext(), maybe we need to filter the nested elements more.

I can give a try with it but I would need some guidance: I'm familiar with this library for less than a week :)

adbar:
Of course, I'll look for potential ways to fix it and give you the necessary info.

adbar:
1. The link in the test sample shouldn't actually be in the output because it's not required in the options and links in tables are often superfluous content. So the tests are working correctly for the setting they're for.
2. `.itertext()` contains element tails, setting `with_tail=False` makes things marginally better but doesn't solve the main problem
3. The best solution is probably to iterate through the element's children (`if len(child) >0` and then `child.iterdescendants()`). You can see how that's done in the `handle_lists()` function above and/or directly use this function `if child.tag == "list"`.

The code is convoluted and badly needs to be simplified, however HTML documents are not as regular as they should on the web so that's why a lot of safeguards have been implemented at different levels in the course of time.

mikhainin:
Thanks - that's helpful!

I updated the implementation it pass our tests and Trafilatura's. Could you take a look once again, please?

adbar:
The walrus operator is not available before Python 3.8 but that's a minor issue.

mikhainin:
Removed the operator

adbar:
The tests pass but on the [benchmark](https://github.com/adbar/trafilatura/blob/master/tests/README.rst) the precision is lower. It could be that tables with too many nested elements contain undesirable text.

mikhainin:
I rebased on the latest master and run `python tests/comparison_small.py`
This branch:
```
nothing
{'true positives': 0, 'false positives': 0, 'true negatives': 2250, 'false negatives': 2236, 'time': 0}
baseline
{'true positives': 1886, 'false positives': 610, 'true negatives': 1640, 'false negatives': 350, 'time': 8.721751689910889}
precision: 0.756 recall: 0.843 accuracy: 0.786 f-score: 0.797
trafilatura
{'true positives': 1991, 'false positives': 183, 'true negatives': 2067, 'false negatives': 245, 'time': 39.170817136764526}
time diff.: 4.49
precision: 0.916 recall: 0.890 accuracy: 0.905 f-score: 0.903
trafilatura + fallback
{'true positives': 2027, 'false positives': 182, 'true negatives': 2068, 'false negatives': 209, 'time': 57.875588178634644}
time diff.: 6.64
precision: 0.918 recall: 0.907 accuracy: 0.913 f-score: 0.912
```

Master:
```
$ python tests/comparison_small.py
number of documents: 750
nothing
{'true positives': 0, 'false positives': 0, 'true negatives': 2250, 'false negatives': 2236, 'time': 0}
baseline
{'true positives': 1886, 'false positives': 610, 'true negatives': 1640, 'false negatives': 350, 'time': 4.424571752548218}
precision: 0.756 recall: 0.843 accuracy: 0.786 f-score: 0.797
trafilatura
{'true positives': 1991, 'false positives': 183, 'true negatives': 2067, 'false negatives': 245, 'time': 18.823368310928345}
time diff.: 4.25
precision: 0.916 recall: 0.890 accuracy: 0.905 f-score: 0.903
trafilatura + fallback
{'true positives': 2027, 'false positives': 182, 'true negatives': 2068, 'false negatives': 209, 'time': 26.65433382987976}
time diff.: 6.02
precision: 0.918 recall: 0.907 accuracy: 0.913 f-score: 0.912
```

If you tell me the right way to run the benchmark, please?

adbar:
I still see a small difference, you need to re-install the package from the branch to see it:
`pip3 install --no-deps -U . ; python3 tests/comparison_small.py`

It's not a big deal, in the worst case your PR would be used when `favor_recall` is activated.

mikhainin:
Yeah in `kath.net-Menschensohn.html`, master version only includes "Mehr zu" header and the current change includes all the links

adbar:
I would then change the condition to `elif child.tag == "list" and favor_recall`.

alroythalus:
web_content = "".join(
        extract(
            web_content,
            include_formatting=True,
            include_tables=True,
            include_comments=False,
            include_links=False,
            output_format="xml",
            favor_recall=True,
            config=config,
        )
    ) 
    
    
   hows the code preserve the indentation of the list items?
   Eg site: https://www.spotify.com/in-en/legal/privacy-policy/
   @adbar @mikhainin

Issue #538, opened 2024-04-03
Make cascade of different content extractors explicit and configurable


adbar:
So far Trafilatura is entwined with a version of readability-lxml, it also uses jusText as fallback before triggering the baseline extraction as last resort. This combination is robust and performs well in the benchmark, however it can be beneficial to refactor the code so as to expose the extractor chain.

The current configuration can be written as follows:
- fast mode: `["trafilatura"]`
- normal mode: `["trafilatura+readability", "justext", "baseline"]`

PR #544, opened 2024-04-04
Respect no_fallback

co-odw:
Problem:
no_fallback is ignored, when len_text < config.getint('DEFAULT', 'MIN_EXTRACTED_SIZE'):

Solution:
respect no_fallback

Not sure if this is desired, but was a little unexpected from my end.


adbar:
@co-odw, Thanks for the feedback, the baseline helps with recall, maybe the name `no_fallback` is debatable here. I'm planning to transition to a modular extraction where the cascade of different extractors is made explicit: #538.

In the meantime I believe this is rather a documentation issue since this behavior is expected (on my side at least).

co-odw:
Sounds good.

PR #587, opened 2024-05-06
Port of is_probably_readerable from mozilla

zirkelc:
Port of [`isProbablyReaderable`](https://github.com/mozilla/readability/blob/main/Readability-readerable.js) from Mozilla.

In the [other PR](https://github.com/adbar/trafilatura/blob/8e6afae6e0b933f6503572484af3606ec2a13698/trafilatura/readability_utils.py), I have used BeautifulSoup and in this I'm using the LXML functions to select the nodes. For whatever reason, there seems to be a difference in the result when running it on the evaluation dataset. I already debugged it, but couldn't figure it out yet. The XPath selector selects a different set of elements than the CSS selector. Maybe you can spot a difference that I'm overlooking? 

I didn't include any tests yet. The official function from Mozilla uses the following test cases to test for `isProbablyReaderable=False` (see attribute `"readerable": false`):
- https://github.dev/mozilla/readability/blob/main/test/test-pages/nytimes-1/expected-metadata.json
- https://github.dev/mozilla/readability/blob/main/test/test-pages/nytimes-1/expected-metadata.json
- https://github.dev/mozilla/readability/blob/main/test/test-pages/nytimes-1/expected-metadata.json
- https://github.dev/mozilla/readability/blob/main/test/test-pages/nytimes-1/expected-metadata.json

I can copy these files if you think it makes sense. 

Please let me know if there is anything to be changed or added.

### DIFF
--- a/trafilatura/readability_lxml.py
+++ b/trafilatura/readability_lxml.py
@@ -493,7 +493,7 @@ def is_probably_readerable(html, options={}):
     min_score = options.get("min_score", 20)
     visibility_checker = options.get("visibility_checker", is_node_visible)
 
-    nodes = doc.xpath(".//p | .//pre | .//article | //div[br]")
+    nodes = doc.xpath(".//p | .//pre | .//article | .//div[br]")
     # br_nodes = doc.xpath(".//div[br]")
 
     node_set = set(nodes)
--- a/trafilatura/readability_lxml.py
+++ b/trafilatura/readability_lxml.py
@@ -491,11 +491,10 @@ def is_probably_readerable(
     """
     doc = load_html(html)
 
-    nodes = doc.xpath(".//p | .//pre | .//article")
+    nodes = doc.xpath(".//p | .//pre | .//article | .//div[.//br]")
+    # br_nodes = doc.xpath(".//div[br]")
 
-    # Adding divs that directly contain <br> tags
-    br_nodes = doc.xpath(".//div[br]")
-    node_set = set(nodes + br_nodes)
+    node_set = set(nodes)
 
     score = 0
     for node in node_set:
--- a/trafilatura/readability_lxml.py
+++ b/trafilatura/readability_lxml.py
@@ -493,10 +493,12 @@ def is_probably_readerable(html, options={}):
     min_score = options.get("min_score", 20)
     visibility_checker = options.get("visibility_checker", is_node_visible)
 
-    nodes = doc.xpath(".//p | .//pre | .//article | .//div[br]")
-    # br_nodes = doc.xpath(".//div[br]")
-
+    nodes = doc.xpath(".//p | .//pre | .//article")
     node_set = set(nodes)
+
+    br_nodes = doc.xpath(".//div/br")
+    for node in br_nodes:
+        node_set.add(node.getparent())
 
     score = 0
     for node in node_set:
--- a/trafilatura/readability_lxml.py
+++ b/trafilatura/readability_lxml.py
@@ -493,7 +493,7 @@ def is_probably_readerable(html, options={}):
     min_score = options.get("min_score", 20)
     visibility_checker = options.get("visibility_checker", is_node_visible)
 
-    nodes = doc.xpath(".//p | .//pre | .//article | .//div[.//br]")
+    nodes = doc.xpath(".//p | .//pre | .//article | //div[br]")
     # br_nodes = doc.xpath(".//div[br]")
 
     node_set = set(nodes)
--- a/trafilatura/readability_lxml.py
+++ b/trafilatura/readability_lxml.py
@@ -483,13 +483,15 @@ def is_node_visible(node):
     return True
 
 
-def is_probably_readerable(
-    html, min_score=20, min_content_length=140, visibility_checker=is_node_visible
-):
+def is_probably_readerable(html, options={}):
     """
     Decides whether or not the document is reader-able without parsing the whole thing.
     """
     doc = load_html(html)
+
+    min_content_length = options.get("min_content_length", 140)
+    min_score = options.get("min_score", 20)
+    visibility_checker = options.get("visibility_checker", is_node_visible)
 
     nodes = doc.xpath(".//p | .//pre | .//article | .//div[.//br]")
     # br_nodes = doc.xpath(".//div[br]")
--- a/trafilatura/readability_lxml.py
+++ b/trafilatura/readability_lxml.py
@@ -470,11 +470,10 @@ def is_node_visible(node):
     """
     Checks if the node is visible by considering style, attributes, and class.
     """
-    style = node.get("style")
 
-    if style and DISPLAY_NONE.search(style) is not None:
+    if "style" in node.attrib and DISPLAY_NONE.search(node.get("style")):
         return False
-    if node.get("hidden") is not None:
+    if "hidden" in node.attrib:
         return False
     if node.get("aria-hidden") == "true" and "fallback-image" not in node.get(
         "class", ""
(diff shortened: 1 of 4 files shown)


adbar:
Hi @zirkelc, thanks for the PR, the code looks good.
Yes, please add basic tests for the functions you added. At best in a new function in `tests/unit_tests.py` or in a new test script. The Mozilla tests would be fine, it could also be a HTML snippet if you can isolate the right parts, as you prefer.

It could be that the combination LXML+XPath works slightly differently than your original setting (BS4+CSS selector), could you please tell me where exactly the difference occurs?

zirkelc:
Hi @adbar 

thanks for the feedback! I ported the [unit tests](https://github.com/mozilla/readability/blob/main/test/test-isProbablyReaderable.js) from mozilla and one [test page ](https://github.com/mozilla/readability/blob/main/test/test-pages/mozilla-2/source.html). Let me know if I should add or change something.

Regarding the XPath vs CSS selector: I will debug it once for to see if I can isolate the discrepancy.

adbar:
A test is failing because you added a file to the resource directory, could you please fix this?

`assert 10 <= len(list(cli_utils.generate_filelist(RESOURCES_DIR))) <= 20` in `cli_tests.py::test_input_type()`

zirkelc:
Sorry, I should have ran all tests. I increased the assertions to 21. Is it okay to put the file into resources or is another folder better suited?

adbar:
It works, thanks. You should now add tests for the lines not covered by the HTML file (see coverage report in "Files changed" tab).

trafilatura/readability_lxml.py
```diff
@@ -440,3 +443,82 @@ def sanitize(self, node, candidates):
...
+    """
+    Decides whether or not the document is reader-able without parsing the whole thing.
+    """
+    doc = load_html(html)
+
+    min_content_length = options.get("min_content_length", 140)
+    min_score = options.get("min_score", 20)
+    visibility_checker = options.get("visibility_checker", is_node_visible)
+
+    nodes = doc.xpath(".//p | .//pre | .//article | .//div[.//br]")
```
adbar, line 496: I'm not sure what `.//div[.//br]` does here, maybe you mean `.//div//br`?

zirkelc: I tried to convert the CSS selector `"div > br"` to XPath, but it probably didn't work out. I think the correct XPath is `//div[br]`?

zirkelc:
I added some more tests which should cover the lines. I also changed the XPath and added test for it

trafilatura/readability_lxml.py
adbar: Then I'd say it's `.//div//br` (br within div), the square brackets are for attributes if I'm not mistaken.

adbar:
There is something wrong with the syntax, see flake error message.

trafilatura/readability_lxml.py
zirkelc: I'm not really fluent with XPath, so I asked ChatGPT on the difference:

- `.//div[br]`: This XPath selects all `<div>` elements that contain at least one `<br>` element directly inside them. The `[br]` in the XPath is a condition that filters the `<div>` elements to only those that have a child `<br>`. It does not select the `<br>` elements themselves, but rather the `<div>` elements that meet this condition.

- `.//div//br`: This XPath selects all `<br>` elements that are descendants of a `<div>` element, regardless of how deep they are nested within the `<div>`. This could include `<br>` elements that are directly inside a `<div>`, as well as `<br>` elements that are nested further inside other child elements of the `<div>`. This XPath essentially traverses all levels of nested elements within each `<div>` to find `<br>` elements.

The original CSS selector was a [child selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) with `div > br`:
> The child combinator matches only those elements matched by the second selector that are the direct children of elements matched by the first.

The square brackets are predicates that contain normal syntax. The attribute selector uses an `@` symbol like `.//div//[@attribute="value"]`. 

Here are some example s that also use this syntax:
> /PROJECT[STATUS]: Selects all projects that have a STATUS child element.
https://docs.oracle.com/javase%2Ftutorial%2F/jaxp/xslt/xpath.html

So I would say `.//div[br]` should be the right conversion from css `div > br`?

adbar: For direct children I would use `//div/br`, maybe there are two different ways to write it.

I wouldn't trust GPT models on questions like this. Let's write a test for it if it's not already the case, that will settle the discussion.

adbar:
I assume the problem is related to nested quotation marks, could you check again?

trafilatura/readability_lxml.py
zirkelc: Here is result:
- `.//div[br]`: http://xpather.com/x39v3Sab
- `//div/br`: http://xpather.com/qgAWnSEC
- `.//div//br`: http://xpather.com/zL4D5mR5

So `//div/br` selects the `<br>` elements while `.//div[br]` selects the `<div>` elements. I also noticed my mistake when porting from JS to Python and why I had problems with the XPath selector in the first place:

https://github.com/mozilla/readability/blob/97db40ba035a2de5e42d1ac7437893cf0da31d76/Readability-readerable.js#L64-L71

This CSS selector selects the `<br>` but it adds the parent element (`<div`>) to the nodes set. I kept overlooking this.

Thanks for be persistent on that!

trafilatura/readability_lxml.py
```diff
@@ -440,3 +443,84 @@ def sanitize(self, node, candidates):
...
+}
+
+DISPLAY_NONE = re.compile(r"display:\s*none", re.I)
+
+
+def is_node_visible(node):
+    """
+    Checks if the node is visible by considering style, attributes, and class.
+    """
+    style = node.get("style")
```
adbar, line 473: You can rewrite the code to make it shorter:
- `if "style" in node.attrib and DISPLAY_NONE.search(node.get("style")):` ...
- same for "hidden" below

trafilatura/readability_lxml.py
```diff
@@ -440,3 +443,84 @@ def sanitize(self, node, candidates):
...
+    """
+    Decides whether or not the document is reader-able without parsing the whole thing.
+    """
+    doc = load_html(html)
+
+    min_content_length = options.get("min_content_length", 140)
+    min_score = options.get("min_score", 20)
+    visibility_checker = options.get("visibility_checker", is_node_visible)
+
+    nodes = doc.xpath(".//p | .//pre | .//article")
```
adbar, line 496: We don't need an intermediary variable "nodes" right? You can make a set out of the xpath return values directly.

Same for "br_nodes" below.

trafilatura/readability_lxml.py
```diff
@@ -440,3 +443,84 @@ def sanitize(self, node, candidates):
...
+        class_and_id = f"{node.get('class', '')} {node.get('id', '')}"
+        if REGEXPS["unlikelyCandidates"].search(class_and_id) and not REGEXPS[
+            "okMaybeItsACandidate"
+        ].search(class_and_id):
+            continue
+
+        if node.xpath("./parent::li/p"):
+            continue
+
+        text_content = node.text_content().strip()
```
adbar, line 517: You can skip the definition of "text_content" since it's not re-used.

adbar (changes requested):

adbar:
Thanks, now the tests pass. I entered a series of minor changes to implement, the PR can soon be merged.

trafilatura/readability_lxml.py
zirkelc: Like this?

zirkelc:
Thanks, please take a look at the changes

adbar (approved):

adbar:
Additional notes:
- The regular expressions used here are slightly different from the legacy ones at the top of the file, probably because they're newer? It would be nice to test if the old ones have to be replaced.
- The additional functions could be documented before a new version comes out.

zirkelc:
> Additional notes:
> 
> * The regular expressions used here are slightly different from the legacy ones at the top of the file, probably because 

Yes, the regex are newer with some additional classes:

https://github.com/mozilla/readability/blob/97db40ba035a2de5e42d1ac7437893cf0da31d76/Readability.js#L123-L126

However, as readability_lxml is based on two other ports of readability and therefore probably out of sync with https://github.com/mozilla/readability/tree/main in general, maybe a completely new port from the JS code would make sense in this case?

they're newer? It would be nice to test if the old ones have to be replaced.

> * The additional functions could be documented before a new version comes out.

Should I add something to the docs?

adbar:
I can take care of the docs before the next release and you can improve on that later if you want.

As you say the readability_lxml module is out of sync but it's not completely different either so it's probably adaptable. If you wish to improve the current state you're very welcome to do so. Bit by bit (maybe easier) or with a complete redesign, as you prefer.

The current version achieves a fine balance between accuracy and precision, so at best changes would need to be tested on the Trafilatura's benchmark (and/or on more comprehensive benchmarks).

Issue #599, opened 2024-05-17
Table markdown syntax incorrect in some cases


naktinis:
This is actually multiple issues with table markdown rendering. Let me know if you'd like me to split it into multiple issues.

Table structure below is taken from wikipedia pages.

Version tested: `1.9.0`

I'll summarize the issues here, but see below for full examples and explanations:
- `<br>` gets deleted but probably should be replaced with a space instead
- New line sometimes gets added before `|`
- `|` completely missing after some rows
- Breaks when `colspan` is used, but could handle it by adding an extra empty cell (append `|` to all lines as many times as colspan in max colspan row requires)
- `<p>` in a cell inserts a new line
- List items in a cell completely disappeared, but could be turned into plain text
- Inserts `---|` in the middle of the table (but `---|` should only appear once after the first row)
- `<wbr>` not handled (should probably be replaced with an empty string)

Extract call used:
```
extract(html_plants, no_fallback=True, output_format='txt', include_tables=True)
```

# Example 1

HTML:
```
<html>
  <table>
    <tbody>
    <tr><th colspan="2">Plants<br>Temporal range: Mesoproterozoic–present Pha. Proterozoic Archean Had.</th></tr>
    <tr><td colspan="2">Angiosperm Desmid Moss Glaucophyta Charophyta Rhodophyta Fern Spirotaenia</td></tr>
    <tr><th colspan="2">Scientific classification</th></tr>
    <tr><td>Domain:</td><td>Eukaryota</td></tr>
    <tr><td>Clade:</td><td>Diaphoretickes</td></tr>
    <tr><td>(unranked):</td> <td>Archaeplastida</td></tr>
    <tr><td>Kingdom:</td><td>Plantae<br>H.F.Copel., 1956</td></tr>
    <tr><th colspan="2">Superdivisions</th></tr>
    <tr><td colspan="2"> <p><i>see text</i> </p> </td></tr>
    <tr><th colspan="2">Synonyms </th></tr>
    <tr><td colspan="2"> <ul><li>Viridiplantae <small>Cavalier-Smith 1981</small><sup>[1]</sup></li> <li>Chlorobionta <small>Jeffrey 1982, emend. Bremer 1985, emend. Lewis and McCourt 2004</small><sup>[2]</sup></li></ul> </td></tr>
    </tbody>
 </table>
</html>
```

Output:
```
PlantsTemporal range: Mesoproterozoic–present Pha. Proterozoic Archean Had.
|
---|
[...]

PR #601, opened 2024-05-17
Markdown table fixes

naktinis:
Contains the following fixes:
- do not add new lines in markdown cells
- markdown tables can only have one header
- add a space after text before a paragraph in a cell
- match maximum cell count on each row
- cells always need to append vertical bars

Partly fixes https://github.com/adbar/trafilatura/issues/599.

### DIFF
--- a/trafilatura/xml.py
+++ b/trafilatura/xml.py
@@ -297,7 +297,7 @@ def process_element(element, returnlist, include_formatting):
         if element.tag in NEWLINE_ELEMS:
             # add line after table head
             if element.tag == "row":
-                max_span = int(element.get("span", 0))
+                max_span = int(element.get("span", 1))
                 cell_count = len(element.xpath(".//cell"))
                 # row ended so draw extra empty cells to match max_span
                 returnlist.append("|" * (max_span - cell_count) + "\n")
--- a/trafilatura/xml.py
+++ b/trafilatura/xml.py
@@ -302,10 +302,13 @@ def process_element(element, returnlist, include_formatting):
     # Process text
 
     # Common elements (Now processes end-tag logic correctly)
+    within_cell = element.xpath("ancestor::cell")
     if element.tag == 'p' and include_formatting:
-        returnlist.append('\n\u2424\n')
+        if not within_cell:
+            returnlist.append('\n\u2424\n')
     elif element.tag in NEWLINE_ELEMS:
-        returnlist.extend([NEWLINE_ELEMS[element.tag], '\n'])
+        if not within_cell:
+            returnlist.extend([NEWLINE_ELEMS[element.tag], '\n'])
     elif element.tag == 'cell':
         returnlist.extend(" | ")
     elif element.tag == 'comments':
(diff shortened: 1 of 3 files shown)


adbar:
@naktinis Thanks for the PR, the code looks good but tests fails in `test_table_processing()` (part of `unit_tests.py`).

Could you please have a look, change the code (or the test if it doesn't work as intended), and add tests there for your PR if necessary?

naktinis:
@adbar I believe I have now fixed the tests and added three more relating to these changes.

adbar:
@naktinis Thanks, it's much better now!

I have a few questions:
- You're introducing a `span` attribute because otherwise it would be difficult to keep track of the total number of columns, right?
- Why does it yield `<rowspan="2">` (without space) and `<row span="7">` in the tests? I would suggest to remove the attribute from the output once it has been used.

trafilatura/xml.py
```diff
@@ -291,21 +295,33 @@ def process_element(element, returnlist, include_formatting):
...
+            # but nothing more to do with other textless elements
+            return
 
     # Process text
 
     # Common elements (Now processes end-tag logic correctly)
+    within_cell = element.xpath("ancestor::cell")
     if element.tag == 'p' and include_formatting:
-        returnlist.append('\n\u2424\n')
+        if not within_cell:
```
adbar, line 320: Maybe rather these conditions or would something go wrong?
`if element.tag == 'p' and include_formatting and not within_cell:`
(same below)

naktinis:
@adbar 
> You're introducing a span attribute because otherwise it would be difficult to keep track of the total number of columns, right?

Yes, so that each row is aware of any other rows having more columns than it does itself.

> Why does it yield <rowspan="2"> (without space) and <row span="7"> in the tests? I would suggest to remove the attribute from the output once it has been used.

Because the previous line (that was already there before my changes) removes spaces:
```
result = processed.replace('\n', '').replace(' ', '')
```

trafilatura/xml.py
naktinis: The idea here was to terminate processing within the condition branch. For example if `element.tag == 'p' and include_formatting` is satsified then I enter the code below and no other `elif/else` branch will ever run.

If I inline it all as `if element.tag == 'p' and include_formatting and not within_cell:` then other branches may still run (which may not be desired).

So in case of `element.tag == 'p' and include_formatting`, we only ever run the couple of lines below and never fall back to any other condition like the `else` statement or other `elif`.

adbar: OK, that makes sense.

adbar:
Then I'd be in favor of removing the attribute from the output:
- somewhere after `for subelement in table_elem.iterdescendants():`
- `del newrow.attrib["span"]` (if it's present)

Could you please implement the change (or a similar one) and update the tests?

naktinis:
@adbar row attribute cleanup should now be done (also updated the tests).

adbar:
Thanks!

PR #603, opened 2024-05-18
fix typos

RainRat:
no functional change


adbar:
Thanks, typos accumulate over time and it's necessary to check everything.

I corrected the evaluation file: the typo was in the original data.

PR #613, opened 2024-06-03
use with_metadata argument as switch

adbar:


fortyfourforty:
It is a breaking change. I lost 30k of title extraction due to this change.
Not to complain, but why make with_metadata=False the default?

For normal users, more data is a good thing as metadata also provides useful information. For advances users, they know how to turn off metadata extraction for speed.

I suggest making the default True. just my 0.2 cents.

PR #658, opened 2024-07-25
table fix: MemoryError & ValueError during conversion to text

### LINKED ISSUE #657: MemoryError from xml parse
`python`: `3.12.0`
`trafilatura`: `1.11.0`

MemoryError happens here https://github.com/adbar/trafilatura/blob/e9921b3724b5fd6219c683b016f89a9b6a79c99c/trafilatura/xml.py#L325
```
# in this case
max_span = 9007199254740991
cell_count = 3
```

```
url = 'https://docs.stripe.com/declines/codes'
resp = trafilatura.fetch_url(url)
content = trafilatura.extract(resp, favor_precision=True, include_images=True, deduplicate=True)
```

In case you can't access this page: https://github.com/Honesty-of-the-Cavernous-Tissue/trafilatura/blob/master/tests/test.html

adbar:
fixes #657.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -1102,6 +1102,10 @@ def test_table_processing():
     result = extract(htmlstring, no_fallback=True, output_format='txt', config=ZERO_CONFIG, include_tables=True)
     assert "a | b | |" in result
 
+    htmlstring = '<html><body><article><table><tr><td span="2">a</td><td>b</td></tr><tr><td>c</td><td>d</td><td>e</td></tr></table></article></body></html>'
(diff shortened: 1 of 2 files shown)


Honesty-of-the-Cavernous-Tissue:
Hi, i think you forget to check `max_span` threshold here https://github.com/adbar/trafilatura/blob/c60395cc3dcaadf5c66a393aeebec447f2dad7a0/trafilatura/xml.py#L326

because i get MemoryError again (😂) on `https://docs.stripe.com/payments/klarna`

adbar:
That's correct, I didn't test the case with table headers, I just opened #665.

PR #678, opened 2024-08-15
Fixed incorrect variable passed to extract_metadata

jpigla:
An incorrect variable (url) was previously passed to the `filecontent` parameter of the **extract_metadata** function, which expects HTML code as a string.

This problem was solved by passing the result of the **fetch_url** function (stored in the variable `downloaded`) to **extract_metadata**::

```python
from trafilatura import fetch_url, extract_metadata

downloaded = fetch_url('https://github.blog/2019-03-29-leader-spotlight-erin-spiceland/')
extract_metadata(downloaded)
```


adbar:
Good catch, thanks!

PR #682, opened 2024-08-27
Add SOCKS Proxy support

gremid:
This PR adds SOCKS proxy support in `trafilatura.downloads`. It works for requests based on `urllib3` as well as `pycurl`, is configured via the environment variable `http_proxy` and tested with SOCKS5 proxies (with and without BASIC authentication).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,17 @@ jobs:
         os: [ubuntu-latest]
         # https://github.com/actions/python-versions/blob/main/versions-manifest.json
         python-version: ["3.9", "3.11"]  # "3.13-dev"
-        env: [{ MINIMAL: "true" }, { MINIMAL: "false" }]
+        env:
+          - MINIMAL: "true"
+          - MINIMAL: "false"
+          - MINIMAL: "true"
+            http_proxy: "socks5://127.0.0.1:1080/"
+          - MINIMAL: "false"
+            http_proxy: "socks5://127.0.0.1:1080/"
+          - MINIMAL: "true"
+            http_proxy: "socks5://user:pass@127.0.0.1:1081/"
+          - MINIMAL: "false"
+            http_proxy: "socks5://user:pass@127.0.0.1:1081/"
         include:
           # custom python versions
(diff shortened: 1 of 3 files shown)


gremid:
@adbar Please let me know, if I can help with documenting the feature.

adbar:
@gremid Could you do something about test coverage? We can see the docs later.

gremid:
I moved the proxy-related tests from the GitHub Action setup to a parametrized pytest fixture. Thus, the coverage of both code paths in the HTTP fetch routines, with and without proxy setup, should be recorded now. The import and parsing of the environment variable `http_proxy` at the top of `trafilatura.downloads` still is not covered.

gremid:
Wrestling with GitHub Actions and environment settings … Now the variable `PROXY_TEST`, defined in the matrix context, should also be available in the [test run](https://github.com/adbar/trafilatura/commit/b74009831424cda3d7fcbcea61d54b5f1558a50f#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR121).

adbar:
Thanks!

andremacola:
I haven't had time to look at Trafilatura for a while, but this PR caught my attention because in the past I had made a patch to add proxies as well. https://github.com/adbar/trafilatura/pull/332/files

@gremid Congratulations on the tests; at the time, I couldn't give it the attention it deserved (sorry @adbar)

A few considerations:

In my view, the proxy shouldn't be used for every request. In this PR, by setting the http_proxy environment variable, Trafilatura assumes it's to be used for everything.
There are use cases where the proxy is only necessary for certain sites. In my opinion, this should be set at the time of the request by passing a boolean value (perhaps in `trafilatura.fetch_url()`).
Another situation is the need for alternating proxies between requests, and an environment variable applied in the current logic limits this approach.
In my PR, you might find some ideas to further improve this functionality: https://github.com/adbar/trafilatura/pull/332/files

These are just suggestions. Again, congratulations on the PR.

adbar:
Thanks for the feedback. We won't work on this right now but I added your idea to the list (#697).

PR #684, opened 2024-08-29
Added prune xpath to spider

felipehertzer:
Hey @adbar,

This pull request adds the `prune_xpath` feature to the Spider function. As part of this update, I've removed the `process_response` to consolidate the pruning logic into a single function.

Please let me know if you have any feedback or suggestions!

Thanks.

### DIFF
--- a/tests/spider_tests.py
+++ b/tests/spider_tests.py
@@ -159,6 +159,14 @@ def test_crawl_logic():
         params = spider.CrawlParameters("xyz")
     assert len(spider.URL_STORE.urldict) == 0
 
+    # empty request
+    params = spider.CrawlParameters("https://example.org")
+    spider.process_response(None, params)
+    assert len(spider.URL_STORE.urldict) == 0
+    assert params.start == params.base == params.ref == "https://example.org"
+    assert params.i == 0 and params.known_num == 0 and params.is_on
+    assert params.lang is None and params.rules is None
+
     # already visited
     params = spider.init_crawl(url, known=[url])
     assert params.base == "https://httpbun.com"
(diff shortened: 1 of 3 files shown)


felipehertzer:
Hey @adbar,

Thanks for the feedback—I understand how the parameters are set up now and have made the requested changes.

I noticed that the code below directly calls `process_links` instead of using the `process_response` function. How would you prefer we fix this? Should we move everything into `probe_alternative_homepage`, or would you rather replace `process_links` with `process_response` as a quick workaround? (process_response, receives a Response instead of htmlstring)

```python
htmlstring, homepage, new_base_url = probe_alternative_homepage(url)
if htmlstring and homepage and new_base_url:
    # register potentially new homepage
    URL_STORE.add_urls([homepage])
    # extract links on homepage
    process_links(htmlstring, params, url=url)
```

adbar:
Hi @felipehertzer, the tests don't pass for all versions, could you check if you can fix them?

What happened to the line `URL_STORE.add_urls([response.url], visited=True)` (in your first commit), maybe it was a good idea?

The code you mention is not an issue because `probe_alternative_homepage()` returns a string, or am I getting it wrong?

felipehertzer:
Hey @adbar,

The tests are currently pending a solution for the `crawl_page` function. Specifically, when `initial` is set to `true`, the `process_response` function does not execute. I have not found an optimal way to run it effectively when `initial` is either `true` or `false`, as the `true` condition returns an `htmlstring`, while the `false` condition returns a `response`

adbar:
I'm not sure how to untangle that right now, if you have an idea feel free to test it.

adbar:
@felipehertzer I think the best way would be to isolate this code in a function and/or to call it in `process_links()`:

```
    if htmlstring and params.prune_xpath is not None:
        if isinstance(params.prune_xpath, str):
            params.prune_xpath = [params.prune_xpath]
        tree = load_html(htmlstring)
        tree = prune_unwanted_nodes(tree, [XPath(x) for x in params.prune_xpath])
        htmlstring = tostring(tree).decode()
```

felipehertzer:
Yes, you are right, it does make way more sense, I will change it

felipehertzer:
Alright, I just updated it.

PR #707, opened 2024-10-02
fix: set options.source before raising error on empty doc tree

### LINKED ISSUE #705: Trafilatura  crashing with `options.source` field never set
I have built a crawler using Trafilatura.

On rare URLs, the crawler triggers a Trafilatura extraction error.

However this extraction error triggers a Trafilatura crash, as it tries to log the `options.source` field which is set nowhere in the library:

https://github.com/adbar/trafilatura/blob/f57ef0b64b4cf96904e377eb012ebb38f097c518/trafilatura/core.py#L247

This line is not the only one. Many logging instructions rely similarly on this non-existent `options.source` field.

This field should be set explcitly everywhere, or the logging shall be simplified everywhere.

dmoklaf:
Made sure the `options` variable is properly backfilled before any exception can be raised

This fixes #705

### DIFF
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -147,11 +147,6 @@ def bare_extraction(filecontent, url=None, no_fallback=False,  # fast=False,
 
     # load data
     try:
-        tree = load_html(filecontent)
-        if tree is None:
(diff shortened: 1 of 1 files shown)


adbar:
Thanks!

PR #711, opened 2024-10-07
Use rst link instead of markdown link in `docs/index.html`

nzw0301:
I found that one link uses a markdown style but it should be rst.


adbar:
That's correct, thanks!

PR #757, opened 2024-12-02
extract image in textnode

unsleepy22:
fix the issue that images within text nodes are not extracted, which is not unusual in web pages. 
i.e.,
```
<p><img src="aa.jpg"></img><p>
```

### DIFF
--- a/trafilatura/htmlprocessing.py
+++ b/trafilatura/htmlprocessing.py
@@ -20,7 +20,7 @@ from .settings import (
     MANUALLY_CLEANED,
     MANUALLY_STRIPPED,
 )
-from .utils import textfilter, trim
+from .utils import textfilter, trim, is_image_element
 from .xml import META_ATTRIBUTES, delete_element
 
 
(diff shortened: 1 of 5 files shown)


trafilatura/utils.py
```diff
@@ -346,6 +346,20 @@ def trim(string: str) -> str:
         return ""
 
 
+def is_image_element(element: _Element):
```
adbar, line 349: The return type should be `-> bool` right?

trafilatura/htmlprocessing.py
```diff
@@ -226,6 +226,8 @@ def handle_textnode(
     preserve_spaces: bool = False,
 ) -> Optional[_Element]:
     "Convert, format, and probe potential text elements."
+    if is_image_element(elem):
```
adbar, line 229: This will slow down processing, is it necessary?

Suggestion:
`if elem.tag == "graphic" and is_image_element(elem):`

adbar:
Hi @unsleepy22, thanks for the PR. I still have questions about it (see above), could you please have a look?

trafilatura/htmlprocessing.py
unsleepy22: ok, fixed

trafilatura/utils.py
unsleepy22: ok, fixed

unsleepy22:
> Hi @unsleepy22, thanks for the PR. I still have questions about it (see above), could you please have a look?

thanks for your comments, fixed and rebased my commits, would you take a look again?

adbar:
LGTM, thanks.

PR #762, opened 2024-12-05
extraction: refine img src url and fix table extraction bugs

unsleepy22:
1. refine img src url, allow for absolute and relative paths
2. fix bugs in table extraction: 1) the last row with colspan is not well formatted in markdown/txt; 2) remove newlines in table cell markdown in some cases

### DIFF
--- a/trafilatura/main_extractor.py
+++ b/trafilatura/main_extractor.py
@@ -8,6 +8,7 @@ import re  # import regex as re
 
 from copy import deepcopy
 from typing import Any, Optional, Tuple, Set, Union
+from urllib.parse import urljoin
 
 from lxml.etree import _Element, Element, SubElement, strip_elements, strip_tags, tostring
 from lxml.html import HtmlElement
@@ -333,7 +334,7 @@ def handle_paragraphs(element: _Element, potential_tags: Set[str], options: Extr
             newsub.text, newsub.tail = processed_child.text, processed_child.tail
 
             if processed_child.tag == 'graphic':
-                image_elem = handle_image(processed_child)
+                image_elem = handle_image(processed_child, options)
                 if image_elem is not None:
                     newsub = image_elem
             processed_element.append(newsub)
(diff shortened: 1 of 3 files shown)


trafilatura/main_extractor.py
```diff
@@ -12,6 +12,8 @@
 from lxml.etree import _Element, Element, SubElement, strip_elements, strip_tags, tostring
 from lxml.html import HtmlElement
 
+from urllib.parse import urljoin
```
adbar, line 15: This is a standard package, ideally it should be imported before external packages like lxml. See the `isort` tool if you want to do this automatically.

trafilatura/main_extractor.py
```diff
@@ -474,7 +483,10 @@ def handle_image(element: Optional[_Element]) -> Optional[_Element]:
     # post-processing: URLs
     src_attr = processed_element.get("src", "")
     if not src_attr.startswith("http"):
-        processed_element.set("src", re.sub(r"^//", "http://", src_attr))
+        if options is not None and options.url is not None:
+            processed_element.set("src", urljoin(options.url, src_attr))
```
adbar, line 487: It would be better to define a variable like `link` and then just write `processed_element.set("src", link)`. It would probably more the code more readable.

adbar:
Great work, the tests show that you definitely improve table processing with this PR.

I just have a few comments to make the code more readable and easier to maintain.

trafilatura/main_extractor.py
unsleepy22: ok, fixed

trafilatura/main_extractor.py
unsleepy22: ok, fixed

unsleepy22:
> Great work, the tests show that you definitely improve table processing with this PR.
> 
> I just have a few comments to make the code more readable and easier to maintain.

All fixed, would you take a look again?

unsleepy22:
> Great work, the tests show that you definitely improve table processing with this PR.
> 
> I just have a few comments to make the code more readable and easier to maintain.

There could be subtle bugs in table extraction as tables sometimes get really complicated, luckily I may encounter enough cases during my work (collecting LLM corpus), so hopefully bugs converge gradually and we may set up enough unit tests or real world test cases.

PR #764, opened 2024-12-07
Fix eager doctype regex matching

UVMvmfee:
This fixes a bug encountered in the wild: repair_faulty_html() tries to match malformed doctype tags assuming the doctype to be followed by a newline. If there actually is no newline then the regex matches too eager, and parts of the document are removed erroneously.

A reduced test case is
`<!DOCTYPE html><html><head></head><body>Foo <br/> Bar</body></html>` which would be reduced to ` Bar</body></html>` without the fix.

This PR adds a reduced test case and tries to fix eagerness by keeping the regex in tag bounderies. pytest shows no regressions.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -140,6 +140,13 @@ def test_input():
         == '<!DOCTYPE html>\n<html lang="en-US">\n<head/>\n<body/>\n</html>'
     )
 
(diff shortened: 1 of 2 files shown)


adbar:
@UVMvmfee That's correct, thanks for the fix.

PR #767, opened 2024-12-13
more robust table extraction

unsleepy22:
Refine table extraction and add more test cases.

### DIFF
--- a/trafilatura/utils.py
+++ b/trafilatura/utils.py
@@ -464,4 +464,4 @@ def copy_attributes(dest_elem: _Element, src_elem: _Element) -> None:
 
 def is_in_table_cell(elem: _Element) -> bool:
     '''Check whether an element is in a table cell'''
-    return elem.xpath('//ancestor::cell')
+    return bool(elem.xpath('//ancestor::cell'))
--- a/trafilatura/utils.py
+++ b/trafilatura/utils.py
@@ -454,3 +454,18 @@ def text_chars_test(string: Optional[str]) -> bool:
     # or not re.search(r'\w', string)
     # return string is not None and len(string) != 0 and not string.isspace()
     return bool(string) and not string.isspace()  # type: ignore[union-attr]
+
+
+def copy_attributes(dest_elem: _Element, src_elem: _Element) -> None:
+    '''Copy attributes from src element to dest element'''
+    for key in src_elem.keys():
+        dest_elem.set(key, src_elem.attrib[key])
+
+
+def is_in_table_cell(elem: _Element) -> bool:
+    '''Check whether an element is in a table cell'''
+    while elem is not None:
+        if elem.tag == 'cell':
+            return True
+        elem = elem.getparent()
+    return False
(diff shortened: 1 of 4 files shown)


trafilatura/utils.py
```diff
@@ -454,3 +454,21 @@ def text_chars_test(string: Optional[str]) -> bool:
     # or not re.search(r'\w', string)
     # return string is not None and len(string) != 0 and not string.isspace()
     return bool(string) and not string.isspace()  # type: ignore[union-attr]
+
+
+def copy_attributes(dest_elem: _Element, src_elem: _Element) -> None:
+    '''Copy attributes from src element to dest element'''
+    for key in src_elem.keys():
```
adbar, line 461: I would rather recommend something like this, I think it should work:
`dest_elem.attrib = src_elem.attrib`

trafilatura/utils.py
```diff
@@ -454,3 +454,21 @@ def text_chars_test(string: Optional[str]) -> bool:
...
+        dest_elem.set(key, src_elem.attrib[key])
+
+
+def is_in_table_cell(elem: _Element) -> bool:
+    '''Check whether an element is in a table cell'''
+    while elem is not None:
```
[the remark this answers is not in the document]
adbar: Maybe even `//ancestor::cell`

adbar:
@unsleepy22 Thanks for the PR, everything looks OK to me but the code could be improved using the suggestions above.

trafilatura/utils.py
unsleepy22: I tried at first, but attrib of _Element is not writable

trafilatura/utils.py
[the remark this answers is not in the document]
unsleepy22: good suggestion, fixed

unsleepy22:
> @unsleepy22 Thanks for the PR, everything looks OK to me but the code could be improved using the suggestions above.

Thanks for your comments, updated accordingly, would you take a look again?

trafilatura/utils.py
```diff
@@ -454,3 +454,14 @@ def text_chars_test(string: Optional[str]) -> bool:
...
+
+def copy_attributes(dest_elem: _Element, src_elem: _Element) -> None:
+    '''Copy attributes from src element to dest element'''
+    for key in src_elem.keys():
+        dest_elem.set(key, src_elem.attrib[key])
+
+
+def is_in_table_cell(elem: _Element) -> bool:
+    '''Check whether an element is in a table cell'''
+    return elem.xpath('//ancestor::cell')
```
adbar, line 467: @unsleepy22 The `.xpath` method returns a list of elements or `[]`, the type annotation is not correct and since we don't need the list it's useless to return it, I suggest using `return bool(elem.xpath(...))`.

unsleepy22: Yes you're right, type check didn't find this one.

unsleepy22: Fixed, could you take a look again?

Issue #769, opened 2024-12-16
Loss format or data when li contains p


ezscode:
Take this url as an example

https://learn.microsoft.com/en-us/azure/search/retrieval-augmented-generation-overview 

***

1、this will ignore li format, turns just like p node

```html
<li>
  <p>Indexing strategies that load and refresh at scale, for all of your content, at the frequency you require.</p>
</li>
```

***

2、can't get all li content 

```html
<ul>
  <li><a href="https://aka.ms/azai/py" data-linktype="external">Python</a></li>
  <li><a href="https://aka.ms/azai/net" data-linktype="external">.NET</a></li>
  <li><a href="https://aka.ms/azai/js" data-linktype="external">JavaScript</a></li>
  <li><a href="https://aka.ms/azai/java" data-linktype="external">Java</a></li>
</ul>
```

***

3、this will igonre format after the first li element 

```html
<ul>
  <li>Azure AI Foundry, <a href="/en-us/azure/ai-studio/concepts/retrieval-augmented-generation" data-linktype="absolute-path">use a vector index and retrieval augmentation</a>.</li>
  <li>Azure OpenAI, <a href="/en-us/azure/ai-services/openai/concepts/use-your-data" data-linktype="absolute-path">use a search index with or without vectors</a>.</li>
  <li>Azure Machine Learning, <a href="/en-us/azure/machine-learning/how-to-create-vector-index" data-linktype="absolute-path">use a search index as a vector store in a prompt flow</a>.</li>
</ul>
```

***

the result is : 

```markdown
- Azure AI Foundry,
[use a vector index and retrieval augmentation](https://learn.microsoft.com/en-us/azure/ai-studio/concepts/retrieval-augmented-generation). - Azure OpenAI,
[use a search index with or without vectors](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/use-your-data). - Azure Machine Learning,
[use a search index as a vector store in a prompt flow](https://learn.microsoft.com/en-us/azure/machine-learning/how-to-create-vector-index).
```

***

I expect below :

```markdown
- Azure AI Foundry,
[use a vector index and retrieval augmentation](https://learn.microsoft.com/en-us/azure/ai-studio/concepts/retrieval-augmented-generation). 
- Azure OpenAI,
[use a search index with or without vectors](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/use-your-data). 
- Azure Machine Learning,
[...]

Issue #553, opened 2024-04-09
Preserve horizontal space in code blocks


mittsommer:
Hello,
thanks for yours continous work on trafilatura
recent when we using trafilatura working on code-text content extraction, wo noticed that the santize func remove all white space \ table even in code block when using txt outpput formating
we think the problem is here preserve_space=False in default
https://github.com/adbar/trafilatura/blob/2c9f20296c1c5ce9a23715a07df5b623f3016b65/trafilatura/xml.py#L315C5-L315C51

adbar:
Do you mean space before the code or space in general? Could you provide a concrete example of code block?

mittsommer:
Guten Tag,
thank you for your replay
we are working on output article with code inside in markdown formating, here is an example 

> 这样在当前目录下就能够生成demo的api服务了。
下图为生成的项目目录结构:
在logic下面的demologic.go编写逻辑
-
```
func (l *DemoLogic) Demo(req *types.Request) (resp *types.Response, err error) {
// todo: add your logic here and delete this line
return &types.Response{
Message: "hello world",
}, nil
}
```

in this case, all white space before the code line in the code block were removed, which is unexpected and not friendly for LLM training

btw. here is another bug (maybe) when extracting inline code block, a redundant '\n' was added after a inline code block 
now result
> #### 1.2、实现`WebMvcConfigurer`
接口，注册拦截器
which is supposed to be
> #### 1.2、实现`WebMvcConfigurer`接口，注册拦截器

thank you

adbar:
Yes, spacing is not necessarily preserved in code blocks, this can be improved.

PR #772, opened 2024-12-24
extraction: improve spacing in item, cell and code blocks

unsleepy22:
Currently preserve_space is forced False in sanitize method when converting xml to text, while for code blocks esp. python code we wish to preserve spaces to keep the original format since space matters in python.

### DIFF
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -91,10 +91,10 @@ def determine_returnstring(document: Document, options: Extractor) -> str:
             header += "---\n"
         else:
             header = ""
-        returnstring = f"{header}{xmltotxt(document.body, options.formatting, options.preserve_space)}"
+        returnstring = f"{header}{xmltotxt(document.body, options.formatting)}"
         if document.commentsbody is not None:
             returnstring = \
-                f"{returnstring}\n{xmltotxt(document.commentsbody, options.formatting, options.preserve_space)}".strip()
+                f"{returnstring}\n{xmltotxt(document.commentsbody, options.formatting)}".strip()
     # normalize Unicode format (defaults to NFC)
     return normalize_unicode(returnstring)
 
@@ -141,7 +141,6 @@ def bare_extraction(
     include_tables: bool = True,
     include_images: bool = False,
     include_formatting: bool = False,
-    preserve_space: bool = False,
     include_links: bool = False,
     deduplicate: bool = False,
     date_extraction_params: Optional[Dict[str, Any]] = None,
@@ -173,7 +172,6 @@ def bare_extraction(
         include_images: Take images into account (experimental).
         include_formatting: Keep structural elements related to formatting
             (present in XML format, converted to markdown otherwise).
-        preserve_space: Preserve space when formatting text.
         include_links: Keep links along with their targets (experimental).
         deduplicate: Remove duplicate segments and documents.
         date_extraction_params: Provide extraction parameters to htmldate as dict().
@@ -208,7 +206,6 @@ def bare_extraction(
             recall=favor_recall,
             comments=include_comments,
             formatting=include_formatting,
-            preserve_space=preserve_space,
             links=include_links,
             images=include_images,
             tables=include_tables,
@@ -365,7 +362,6 @@ def extract(
     include_tables: bool = True,
     include_images: bool = False,
     include_formatting: bool = False,
-    preserve_space: bool = False,
     include_links: bool = False,
     deduplicate: bool = False,
     date_extraction_params: Optional[Dict[str, Any]] = None,
@@ -399,7 +395,6 @@ def extract(
         include_images: Take images into account (experimental).
         include_formatting: Keep structural elements related to formatting
             (only valuable if output_format is set to XML).
-        preserve_space: Preserve space when formatting text.
         include_links: Keep links along with their targets (experimental).
         deduplicate: Remove duplicate segments and documents.
         date_extraction_params: Provide extraction parameters to htmldate as dict().
@@ -433,7 +428,6 @@ def extract(
         include_tables=include_tables,
         include_images=include_images,
         include_formatting=include_formatting,
-        preserve_space=preserve_space,
         include_links=include_links,
         deduplicate=deduplicate,
         date_extraction_params=date_extraction_params,
@@ -463,7 +457,6 @@ def extract_with_metadata(
     include_tables: bool = True,
     include_images: bool = False,
     include_formatting: bool = False,
-    preserve_space: bool = False,
     include_links: bool = False,
     deduplicate: bool = False,
     date_extraction_params: Optional[Dict[str, Any]] = None,
@@ -495,8 +488,7 @@ def extract_with_metadata(
         include_images: Take images into account (experimental).
         include_formatting: Keep structural elements related to formatting
             (only valuable if output_format is set to XML).
-        preserve_space: Preserve space when formatting text.
-        include_links: Keep links along with their targets (experimental).
+=        include_links: Keep links along with their targets (experimental).
         deduplicate: Remove duplicate segments and documents.
         date_extraction_params: Provide extraction parameters to htmldate as dict().
         url_blacklist: Provide a blacklist of URLs as set() to filter out documents.
@@ -524,7 +516,6 @@ def extract_with_metadata(
         include_tables=include_tables,
         include_images=include_images,
         include_formatting=include_formatting,
-        preserve_space=preserve_space,
         include_links=include_links,
         deduplicate=deduplicate,
         date_extraction_params=date_extraction_params,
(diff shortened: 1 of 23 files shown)


adbar:
Hi @unsleepy22, a few comments on the PR:
- I believe this fixes #553, could you please check?
- There is already a `preserve_spaces` parameter in [htmlprocessing/handle_textnode](https://github.com/adbar/trafilatura/blob/91c567c3efc4bd8e637c2820abeb72c71edc73d3/trafilatura/htmlprocessing.py#L226), is it necessary to use a new argument across multiple functions?
- This behavior could actually become the default if there are no adverse effects, it would be easier for maintenance

Last, if you have several PRs in store about unrelated parts of the code it would be best to send them all at once, I don't plan to maintain the software on a day-to-day basis, we could both save time by grouping requests.

unsleepy22:
> Hi @unsleepy22, a few comments on the PR:
> 
> * I believe this fixes [Preserve horizontal space in code blocks #553](https://github.com/adbar/trafilatura/issues/553), could you please check?
> * There is already a `preserve_spaces` parameter in [htmlprocessing/handle_textnode](https://github.com/adbar/trafilatura/blob/91c567c3efc4bd8e637c2820abeb72c71edc73d3/trafilatura/htmlprocessing.py#L226), is it necessary to use a new argument across multiple functions?
> * This behavior could actually become the default if there are no adverse effects, it would be easier for maintenance
> 
> Last, if you have several PRs in store about unrelated parts of the code it would be best to send them all at once, I don't plan to maintain the software on a day-to-day basis, we could both save time by grouping requests.

Sorry for late response as I have been on vacation for a few days.

Yes, this fixes [Preserve horizontal space in code blocks #553](https://github.com/adbar/trafilatura/issues/553).
And also I added code to fix https://github.com/adbar/trafilatura/issues/769.

The `preserve_spaces` option does not work globally in [htmlprocessing/handle_textnode], and to keep backward-compatibility, I leave `preserve_spaces` to False by default.

Another change I'd like to make but needs discussion is adding another option like `include_metadata_in_text` to preserve/remove metadata dict string from text, but this may cause a bit confusion since we have 3 options regarding metadata, and, again, just for backward-compatibility. Another way to do this is change the default behavior of `with_metadata` option, which might introduce breaking changes. What do you think?

adbar:
- If `preserve_spaces` is nearly always set to false, wouldn't it be easier to just remove the parameter and consider that space should be preserved in certain elements only like code blocks?
- I'm not sure about `with_metadata`, what do you want to achieve by doing this? Metadata is already added in TXT/MD format when `with_metadata=True` or is there a bug?

unsleepy22:
> * If `preserve_spaces` is nearly always set to false, wouldn't it be easier to just remove the parameter and consider that space should be preserved in certain elements only like code blocks?
It's ok with me to make this change, I'll change my code a bit and fix a few UTs.

> * I'm not sure about `with_metadata`, what do you want to achieve by doing this? Metadata is already added in TXT/MD format when `with_metadata=True` or is there a bug?
Currently if `with_metadata` is set to True, the metadata dict string will be prepended before text, what I want is to exclude metadata dict string from text and returned within Document obj.

adbar:
OK then, you can change the code of  `preserve_spaces` to avoid adding another parameter.

Concerning `with_metadata` this is expected result when people use `extract()`. To get metadata separately from the text you could just use `bare_extraction()`. The text is then in `.raw_text`, the title in `.title` etc.
If this is not enough please open an issue with an example of the output you want to get.

adbar:
Hi @unsleepy22, are you available to make the final changes to the PR?

unsleepy22:
> Hi @unsleepy22, are you available to make the final changes to the PR?

Sorry it's been quite a while, got a bit busy on work stuff (and a long Chinese Spring Festival vacation). I think it'll take a few more days to fix the changes.

adbar:
That's fine, 蛇年祝福 !

unsleepy22:
This PR fixes #769  and #553 

Also I forced a space after highlighted text like **, *, etc. for better compatibility with various markdown editors as some md editor will fail to render for text like `**hello,**world`.
This PR introduces a minor defect that it adds an extra space in table cells which seems a bit weird, though there's no side effect in md rendering. I'm still trying to find a better way to tackle this.

BTW, #776  looks more elegant in handling code blocks, if this also fixes #553 , I think my PR would be a duplicate fix? I'm OK to merge #776 and simplify my PR.

adbar:
OK then, I will merge #776 and then you can simplify this PR.

I don't really like the idea of adding a space but if it's a common way to write markdown then why not.

unsleepy22:
Checked that #776 does not fully fix #553 so I just merged master and resolved conflicts. @adbar would you take a look?

adbar:
@unsleepy22 Everything looks good but can you please remove the additional spacing? Or do you have a format specification for this? I checked Markdown and Commonmark, the current syntax in master is correct.

unsleepy22:
> @unsleepy22 Everything looks good but can you please remove the additional spacing? Or do you have a format specification for this? I checked Markdown and Commonmark, the current syntax in master is correct.

OK all fixed, would you take a look again?

trafilatura/core.py
```diff
@@ -487,7 +488,7 @@ def extract_with_metadata(
         include_images: Take images into account (experimental).
         include_formatting: Keep structural elements related to formatting
             (only valuable if output_format is set to XML).
-        include_links: Keep links along with their targets (experimental).
+=        include_links: Keep links along with their targets (experimental).
```
adbar, line 491: This can be removed ,right?

adbar:
@unsleepy22 The PR looks good. however I just ran tests on the benchmark and the extraction is really slow after your PRs, it gets even slower after this one. I think we need to fix this.

unsleepy22:
> @unsleepy22 The PR looks good. however I just ran tests on the benchmark and the extraction is really slow after your PRs, it gets even slower after this one. I think we need to fix this.

OK I'll see to it.

adbar:
Thanks, this is much better but the fast mode (without extraction fallbacks) is still slower than the full mode. Could you do something about that?

unsleepy22:
> Thanks, this is much better but the fast mode (without extraction fallbacks) is still slower than the full mode. Could you do something about that?
> 
> Edit: The fast mode is not only slower, now the results are the same as in full mode, so I believe extraction steps happen which shouldn't be there.

Do you mean fast mode in master branch is faster than fast mode in my PR? I checked both on my M4 max and MBP 2019, both show no significant performance gap (average on 30 runs, less than 30ms performance gap on MBP out of overall avg 1800ms and less than 5ms gap on M4 max out of overall avg 300ms).

I ran another test which shows that fast mode produces different result with full mode, did I miss some point?
```
    htmlstring = ('<html><body>'
                  '<table><tr><td>aa</td><td>bb</td></tr><tr><td>aa</td><td>bb</td></tr></table>'
                  '<table><tr><td>aa</td><td>bb</td></tr><tr><td>aa</td><td>bb</td></tr></table>'
                  '<div><details><summary>Epcot Center</summary><p>Epcot is a theme park at Walt Disney World Resort featuring exciting attractions, international pavilions, award-winning fireworks and seasonal special events.</p></details>'
                  '</div><div>aaa<br/>bbb<br/>ccc</div></body></html>')
    print('\n')
    print(extract(htmlstring, fast=True, config=ZERO_CONFIG))
    print(extract(htmlstring, fast=False, config=ZERO_CONFIG))
```

trafilatura/core.py
adbar: @unsleepy22 Could you please remove it if necessary?

adbar:
@unsleepy22 Yes, the problem is now that the "fast" mode isn't faster than the other anymore (somewhere after your first PRs). Not a huge deal but something we could address in another PR.

unsleepy22:
> @unsleepy22 Yes, the problem is now that the "fast" mode isn't faster than the other anymore (somewhere after your first PRs). Not a huge deal but something we could address in another PR.

OK, I see your point, I'll look into this.

trafilatura/core.py
unsleepy22: OK, maybe mis-typed this.

unsleepy22:
> @unsleepy22 Yes, the problem is now that the "fast" mode isn't faster than the other anymore (somewhere after your first PRs). Not a huge deal but something we could address in another PR.

I run comparison_small.py locally, looks it always costs 20+ secs in whichever branch (v1.2.2, v2.0.0, current master, my PR).
The main issue is that in current PR, fast mode is not a lot faster than full mode, am I understanding correctly?

trafilatura/utils.py
```diff
@@ -3,6 +3,7 @@
 Module bundling functions related to HTML and text processing,
 content filtering and language detection.
 """
+from wsgiref.util import request_uri
```
adbar, line 6: Is this needed?

adbar:
@unsleepy22 Yes, it has been like this since one of your recent PRs, I'm not sure which, we would have to test.

Please get rid of the superfluous lines so that I can merge this PR, you can solve the fast/full issue in another PR if you find a way.

trafilatura/utils.py
unsleepy22: OK, removed

brycepg:
I'd really like to see this feature present, please merge :)

Issue #782, opened 2025-02-03
"Chinese dot" breaks the fingerprint method


reinoldus:
Hi everyone,

I noticed that if this "chinese dot" is included in the fingerprint hash "。" then the hash is always "ffffffffffffffff"

```
from trafilatura.deduplication import content_fingerprint, Simhash
content_a = Simhash("""行政長官岑浩""")


print(Simhash("""行政長官岑浩""").to_hex()) # 13dd8c82d4634a48
print(Simhash("""欢迎与我交流""").to_hex()) # 58429793861fa351
print(Simhash("""行政長官岑浩。""").to_hex()) #ffffffffffffffff
print(Simhash("""欢迎与我交流。""").to_hex()) #ffffffffffffffff
```

I hope the encoding is not all messed up.

reinoldus:
I did some more debugging and it seems like the sample_tokens function returns an empty string, because due to the "chinese dot" the check:

```
        if token.isalnum():
            tokens.append(token)
```

fails, because the chinese dot is not an alnum character.

We have a couple of options here:

Use the "regex" lib to remove unicode punctuation

```
    import regex  
    strip_punct = regex.compile(r'\p{P}+')
    token = strip_punct.sub('', token)
```

or this approach:

```
import unicodedata
import sys

tbl = dict.fromkeys(i for i in range(sys.maxunicode)
                      if unicodedata.category(chr(i)).startswith('P'))
def remove_punctuation(text):
    return text.translate(tbl)
```


Or for this particular case we could just append the whole text if no tokens have been found (probably the worst option)

```
    if not tokens:
        tokens.append(inputstring)
```


Let me know if any of those options would work for you and I will open a PR.

adbar:
Hi @reinoldus, thanks for the suggestions, it actually amounts to a tokenization issue.

I believe the most efficient approach is to add the character to `string.punctuation` in a global variable and then to replace this line in the deduplication module (just above):
`token = token.strip(string.punctuation)`

Feel free to draft a pull request on it, I could review and integrate it.

PR #783, opened 2025-02-05
Fixing tokenizer not stripping "Ideographic Full Stop" (chinese full …

reinoldus:
Continuing the discussion from #782 

After some deeper research I think the Unicode category fix would be more appropriate, because it results generally in better tokenization for mandarin. Also I moved the replacing of the characters out of the for-loop to be able to replace punctuation with blanks:

For this text (copied from news.cn):

```
  会谈后，两国元首共同签署《中华人民共和国和吉尔吉斯共和国关于深化新时代全面战略伙伴关系的联合声明》，见证签署共建“一带一路”合作规划以及外交、经贸、农业等领域多项合作文件。
```

The tokenizer in this PR produces:
```
['会谈后', '两国元首共同签署', '中华人民共和国和吉尔吉斯共和国关于深化新时代全面战略伙伴关系的联合声明', '见证签署共建', '一带一路', '合作规划以及外交', '经贸', '农业等领域多项合作文件']
```

Where the current one produces:

```
[]
```

Just adding the "Ideographic Full Stop" as punctuation would also result in `[]`.

The only downside of doing this change is that the fingerprints might change quite significantly, maybe we can hide this behind a feature flag to give people time to update their fingerprints?

Here is some code to verify:

```
import unicodedata
import string
from typing import List


def sample_tokens_CURRENT(inputstring: str, length: int = 64) -> List[str]:
    """Split input into list of tokens and adjust length threshold to make sure
    there is enough data."""
    tokens = []
    for token in inputstring.split():
        token = token.strip(string.punctuation)
        if token.isalnum():
            tokens.append(token)
    sample = []
    for i in range(4, -1, -1):
        sample = [t for t in tokens if len(t) > i]
        if len(sample) >= length / 2:
            return sample
    return sample


################# SAMPLE TOKEN EASY FIX


def sample_tokens_EASY_FIX(inputstring: str, length: int = 64) -> List[str]:
    """Split input into list of tokens and adjust length threshold to make sure
    there is enough data."""
    tokens = []
    for token in inputstring.split():
        token = token.strip(string.punctuation + "。")
        if token.isalnum():
            tokens.append(token)
    sample = []
    for i in range(4, -1, -1):
        sample = [t for t in tokens if len(t) > i]
        if len(sample) >= length / 2:
            return sample
    return sample


################## SAMPLE TOKEN PROPOSE

def strip_all_punctuation(text: str) -> str:
    """Replace all Unicode punctuation characters with spaces."""
    cleaned_chars = []

    for char in text:
        # Check if character is in any Unicode punctuation category
        is_punctuation = unicodedata.category(char).startswith('P')

        # Replace punctuation with space, otherwise keep character
        cleaned_char = ' ' if is_punctuation else char
        cleaned_chars.append(cleaned_char)

    return ''.join(cleaned_chars)


def sample_tokens_unicode_fix(inputstring: str, length: int = 64) -> list[str]:
    """Split input into list of tokens and adjust length threshold to make sure
    there is enough data."""
    tokens = []

    # replace punctuation with blanks beforehand to handle none latin languages better
    inputstring = strip_all_punctuation(inputstring)

    for token in inputstring.split():
        if token.isalnum():
            tokens.append(token)
    sample = []
    for i in range(4, -1, -1):
        sample = [t for t in tokens if len(t) > i]
        if len(sample) >= length / 2:
            return sample
    return sample


# this chinese text was copied from news.cn

from trafilatura import deduplication

## Mix of languages
text = "Hello,World! こんにちは。नमस्ते! ¡Hola!"
print(sample_tokens_CURRENT(text))  # Returns: []
print(sample_tokens_EASY_FIX(text))  # Returns: []
print(sample_tokens_unicode_fix(text))  # Returns: ['Hello', 'World', 'こんにちは', 'Hola']

print("#" * 100)

# Mandarin word
mandarin_word = "行政長官岑浩。"
print(sample_tokens_CURRENT(mandarin_word)) 
print(sample_tokens_EASY_FIX(mandarin_word))
print(sample_tokens_unicode_fix(mandarin_word)) 

print("#" * 100)
# Chinese text
text_chinese = "  会谈后，两国元首共同签署《中华人民共和国和吉尔吉斯共和国关于深化新时代全面战略伙伴关系的联合声明》，见证签署共建“一带一路”合作规划以及外交、经贸、农业等领域多项合作文件。"

print(sample_tokens_CURRENT(text_chinese))  # Returns xxx
print(sample_tokens_EASY_FIX(text_chinese))  # Returns: []
print(sample_tokens_unicode_fix(
    text_chinese))  # Returns: ['会谈后', '两国元首共同签署', '中华人民共和国和吉尔吉斯共和国关于深化新时代全面战略伙伴关系的联合声明', '见证签署共建', '一带一路', '合作规划以及外交', '经贸', '农业等领域多项合作文件']
```

Output:

```
[]
[]
['Hello', 'World', 'こんにちは', 'Hola']
#############################################
[]
['行政長官岑浩']
['行政長官岑浩']
#############################################
[]
[]
['会谈后', '两国元首共同签署', '中华人民共和国和吉尔吉斯共和国关于深化新时代全面战略伙伴关系的联合声明', '见证签署共建', '一带一路', '合作规划以及外交', '经贸', '农业等领域多项合作文件']
```

### DIFF
--- a/trafilatura/deduplication.py
+++ b/trafilatura/deduplication.py
@@ -35,24 +35,26 @@ def is_similar_domain(reference: str, new_string: str, threshold: float = 0.5) -
     return SequenceMatcher(None, reference, new_string).ratio() >= threshold
 
 
+def _get_sample_by_length(tokens: List[str], target_length: int) -> List[str]:
+    """Helper function to get a sample of tokens based on length criteria."""
+    sample = []
+    for i in range(4, -1, -1):
+        sample = [t for t in tokens if len(t) > i]
+        if len(sample) >= target_length / 2:
+            return sample
+    return sample
+
+
 def sample_tokens_fallback(inputstring: str, length: int = 64) -> List[str]:
     """
     This fallback implementation is used when the primary sample_tokens function
     generates an empty token list. This is mostly relevant for languages like
     mandarin where none latin-based punctuation is used e.g.: 。
     """
-    tokens = []
     # Replace all punctuation with spaces using translation table
     clean_text = inputstring.translate(PUNCT_TBL)
-    for token in clean_text.split():
-        if token.isalnum():
-            tokens.append(token)
-    sample = []
-    for i in range(4, -1, -1):
-        sample = [t for t in tokens if len(t) > i]
-        if len(sample) >= length / 2:
-            return sample
-    return sample
+    tokens = [t for t in clean_text.split() if t.isalnum()]
+    return _get_sample_by_length(tokens, length)
 
 
 def sample_tokens(inputstring: str, length: int = 64) -> List[str]:
@@ -63,11 +65,8 @@ def sample_tokens(inputstring: str, length: int = 64) -> List[str]:
         token = token.strip(string.punctuation)
         if token.isalnum():
             tokens.append(token)
-    sample = []
-    for i in range(4, -1, -1):
-        sample = [t for t in tokens if len(t) > i]
-        if len(sample) >= length / 2:
-            return sample
+
+    sample = _get_sample_by_length(tokens, length)
 
     if len(sample) == 0:
         return sample_tokens_fallback(inputstring, length)
@@ -87,10 +86,10 @@ class Simhash:
     __slots__ = ["hash", "length"]
 
     def __init__(
-        self,
-        inputstring: str = "",
-        length: int = 64,
-        existing_hash: Optional[str] = None,
+            self,
+            inputstring: str = "",
+            length: int = 64,
+            existing_hash: Optional[str] = None,
     ) -> None:
         "Store length and existing or new hash."
         self.length = length
@@ -114,7 +113,7 @@ class Simhash:
         #    return -2
         # return x
 
-    @lru_cache(maxsize=2**14)
+    @lru_cache(maxsize=2 ** 14)
     def _vector_to_add(self, token: str) -> List[int]:
         "Create vector to add to the existing string vector"
         return [1 if self._hash(token) & (1 << i) else -1 for i in range(self.length)]
--- a/trafilatura/deduplication.py
+++ b/trafilatura/deduplication.py
@@ -4,8 +4,8 @@
 # 3.11+: from typing import Self
 
 import re
+import sys
 import string
-
 from difflib import SequenceMatcher
 from functools import lru_cache
 from hashlib import blake2b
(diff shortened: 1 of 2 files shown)


adbar:
Hi @reinoldus, thanks for the detailed PR! Everything works and your logic makes perfect sense but I have concerns about existing hashes (as you say) and overall speed:

- I suggest to implement your logic as a fallback, i.e. if to only activate it if nothing is found.
- Writing the new function like this should be more efficient:
`[' ' if unicodedata.category(c).startswith('P') else c for c in text]`
Using `str.maketrans` would also be an option but I'm not sure about the character range.

adbar:
The Python lexer provides a test with `str.isidentifier()`, this would deal with Unicode chars, here a small demo:

```
>>> "Test".isidentifier()
True
>>> "Test.".isidentifier()
False
>>> "".join([c for c in "  会谈后，" if c.isidentifier()])
'会谈后'
```

adbar:
@reinoldus Let's go for translate then, thanks for the tests, that's quite helpful to decide.

Just two things before merging the PR:
1.  The code can be simplified:
    - `tokens = [t for t in clean_text.split() if t.isalnum()]` in your new function
    - the sampling loop is now present twice, it could be a function
2. Please add a few cases of yours under `tests/deduplication_tests.py`, it will be good for regression tests

trafilatura/deduplication.py
```diff
@@ -4,25 +4,28 @@
...
 
 from .settings import LRU_SIZE
 from .utils import trim
 
-
 STRIP_EXTENSION = re.compile(r"\.[^/?#]{2,63}$")
 
 BIN_COUNT_FUNC = getattr(int, "bit_count", lambda x: bin(x).count("1"))
 
+PUNCT_TBL = dict.fromkeys((i for i in range(sys.maxunicode)
```
adbar, line 26: `PUNCT_TBL = {i: ' ' for i in range(0x10FFFF) if unicodedata.category(chr(i)).startswith('P')}`

This is simpler and entering the maximum Unicode codepoint manually removes the need for sys.

adbar:
Hi @reinoldus, we're nearly there, see comment above. I also believe it would be best to make your dict static with [str.maketrans](https://docs.python.org/3/library/stdtypes.html#str.maketrans) on the same line. It probably makes `str.translate` more efficient, otherwise the type has to be converted.

adbar:
Great, thanks.

I don't agree on the speed, look at the maximum values. There are cases where it's a lot better with a static variable. Now we don't need to check why :+1:

PR #791, opened 2025-03-15
Optimize PUNCT_TBL creation 

deedy5:
```python
from time import monotonic
import unicodedata

t0 = monotonic()
PUNCT_TBL = str.maketrans({i: " " for i in range(0x10FFFF) if unicodedata.category(chr(i)).startswith("P")})
print(monotonic() - t0)

t0 = monotonic()
PUNCT_TBL2 = str.maketrans({i: " " for i in range(0x10FFFF) if unicodedata.category(chr(i))[0] == "P"})
print(monotonic() - t0)

print(PUNCT_TBL == PUNCT_TBL2)
```
```python
0.2510066920076497
0.14058679199661128
True
```

### DIFF
--- a/trafilatura/deduplication.py
+++ b/trafilatura/deduplication.py
@@ -22,7 +22,7 @@ STRIP_EXTENSION = re.compile(r"\.[^/?#]{2,63}$")
 
 BIN_COUNT_FUNC = getattr(int, "bit_count", lambda x: bin(x).count("1"))
 
-PUNCT_TBL = str.maketrans({i: ' ' for i in range(0x10FFFF) if unicodedata.category(chr(i)).startswith('P')})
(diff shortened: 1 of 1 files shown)


deedy5:
There is another option (`PUNCT_TBL2`), it is faster but still requires some computation time:
```python
from time import monotonic
import unicodedata

t0 = monotonic()
PUNCT_TBL = str.maketrans(
    {i: " " for i in range(0x10FFFF) if unicodedata.category(chr(i)).startswith("P")}
)
print(monotonic() - t0)

t0 = monotonic()
PUNCT_TBL2 = str.maketrans(
    {ch: " " for ch in map(chr, range(0x110000)) if unicodedata.category(ch)[0] == "P"}
)
print(monotonic() - t0)

print(PUNCT_TBL == PUNCT_TBL2)
```
```python
0.2510066920076497
0.15876542999467347
True
```

PR #802, opened 2025-06-08
Review HTML element list and ensure complete XML conversion coverage

### LINKED ISSUE #720: Review HTML element list and conversion
- [ ] Check if all elements in this list are accounted for:
`https://developer.mozilla.org/en-US/docs/Web/HTML/Element`
- [ ] Check conversion to XML

eyupcanakman:
Closes #720: Review HTML element list and conversion.

Ensured all MDN HTML elements are accounted for and correctly mapped to XML.

- Created `html_elements_reference.py` snapshot including all 95+ MDN elements (modern, legacy, deprecated).
- Preserved explicit semantic mappings (headings → `head`, lists → `list`).
- Added default identity mappings (`tag` → `tag`) ensuring no elements are lost.
- Explicit mappings defined for semantic elements.
- Automatic identity mappings provided as defaults.
- New tests verify complete MDN coverage.

### DIFF
--- a/tests/test_html_elements.py
+++ b/tests/test_html_elements.py
@@ -1,0 +1,56 @@
+"""Test HTML elements coverage against MDN reference."""
+
+from trafilatura.htmlprocessing import HTML_EL_TO_XML_EL
+from trafilatura.html_elements_reference import MDN_ELEMENTS
+
+
+def test_every_mdn_tag_is_mapped() -> None:
+    """Regression guard for GH-720: no MDN element may be forgotten."""
+    missing = sorted(MDN_ELEMENTS - HTML_EL_TO_XML_EL.keys())
+    assert not missing, f"Tags without conversion rule: {missing}"
+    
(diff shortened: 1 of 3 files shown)


adbar:
@eyupcanakman Your PR doesn't change anything in the way documents are processed, I will close it if you don't integrate it into the actual code.

eyupcanakman:
@adbar You're right the code was processing the DOM twice which doesn't make sense. I just pushed a fix for that.

adbar:
@eyupcanakman The last change looks good but I still need to think about the PR. There is a small negative impact on the benchmark.
You get more coverage if you use all the elements but it may the cause of unexpected behavior further down the line as the elements are converted to the output format.

eyupcanakman:
Closing in favor of #847. The original audit-doc scope didn't hold up empirically, so I narrowed it to one focused fix the audit surfaced.

PR #809, opened 2025-09-02
Added support to meta name image

felipehertzer:
Hey @adbar,

How are you?

I noticed that some websites are using the following tag to store the image, and it looks like this one was missing in the metadata:

```html
<meta name="twitter:image" content="https://images.jpg" />
```

Thanks.


adbar:
Hi @felipehertzer, nice to hear from you. I'd ask for a specific test but your change is already covered apparently so it's definitely present in the data!

PR #826, opened 2026-02-23
Fix text inside nobr-tags is dropped

### LINKED ISSUE #825: Text inside nobr-tags is dropped
I noticed that `trafilatura` drops text in `<nobr>` tags.

*Example*: `"Sjätte <nobr>AP-fonden</nobr>"` gets extracted as `"Sjätte "` and not the expected `"Sjätte AP-fonden"`.

kod-kristoff:
This PR fixes the bug #825.

The PR also fixes some error in the tests.

Fixes #825

### DIFF
--- a/tests/cache/data.riksdagen.se.dokument.h2b51.html
+++ b/tests/cache/data.riksdagen.se.dokument.h2b51.html
@@ -1,0 +1,6373 @@
+
+<div style="margin:3px;">
+<STYLE type="text/css">
+
+body {margin-top: 0px;margin-left: 0px;}
+
+#page_1 {position:relative; overflow: hidden;z:b;margin:12px 0px 14px 10px ;padding: 0px;border: none;width: 902px;height: 1040px;}
+
+#page_1 #dimg1z{position:absolute;top:0px;left:0px;z-index:-1;width:710px;height:1040px;}
+#page_1 #dimg1 #img1  {width:710px;height:1040px;}
+
+
+
+
+#page_2 {position:relative; overflow: hidden;z:b;margin:15px 0px 25px 10px ;padding: 0px;border: none;width: 902px;height: 974px;}
+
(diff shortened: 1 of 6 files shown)


adbar:
Hi @kod-kristoff, your PR looks OK but for some reason the tests haven't run.

kod-kristoff:
It seems that you need to approve the workflows
bild

adbar:
You could merge the main branch now to see if things work. I thing adding the examples and the nobr tag to the extraction should be enough, I'm not sure the rest is needed.

kod-kristoff:
I have rebased on master now

adbar:
I think `assert "Features" in res` is `True` now.

kod-kristoff:
> I think `assert "Features" in res` is `True` now.

Yes, I have dropped that commit :smile:

kod-kristoff:
Rebased on master

toxa3500:
Hi @kod-kristoff, thanks for tackling this — I hit the same `<nobr>` bug (#825).

In case it helps get this unstuck: the root cause is just that `nobr` isn't in any of trafilatura's tag lists, so the element gets dropped during conversion and takes its text with it. Adding `"nobr"` to `MANUALLY_STRIPPED` in `settings.py` strips the tag but keeps its text — exactly like the other inline presentational tags (`font`, `big`, `small`). That's a one-line change.

For the regression test, a small inline HTML snippet inside `test_formatting()` (e.g. `"Sjätte <nobr>AP-fonden</nobr> …"` asserting `"AP-fonden"` survives) might be lighter than the full cached fixture — and would sidestep the CI run + the merge conflict this PR is currently hitting.

Happy to help however's useful — thanks again for working on this!

kod-kristoff:
Hi @adbar! 
I have simplified the fix according to @toxa3500 :s suggestion.

Hopefully it will clear all tests now

PR #841, opened 2026-04-16
fix: preserve images when links are included

rmarchei:
## Summary
- Images inside container divs were being deleted by `link_density_test()` when `--links` was enabled, because empty anchor refs (e.g. `<ref target="#img-2"/>`) made the divs look link-heavy. Skip the link-density check when an element has any `<graphic>` descendant.
- Move `<graphic>` elements out of their enclosing `<ref>` in `convert_tags()` to prevent loss during paragraph processing, and drop image-only refs whose leftover children would otherwise trigger the justext fallback path.
- Graphics are lifted out in reverse order so `addnext()` preserves their original document sequence (matters for multi-image links like galleries).

## Test plan
- [x] Added `test_htmlprocessing` assertion covering multi-image link ordering.
- [x] Existing `test_htmlprocessing` suite still passes.


adbar:
Hi @rmarchei, your PR looks OK but for some reason the tests haven't run.

rmarchei:
Hey @adbar,

I pushed an empty ci: retrigger commit as you suggested and the workflows are in `awaiting approval from a maintainer` now. Could you approve them when you get a chance? Thanks!

adbar:
The CI was failing but the main branch works again now, you'd have to merge it here in order to run the tests properly.

adbar:
LGTM, thanks.

PR #847, opened 2026-05-09
fix: clean fencedframe element during extraction

eyupcanakman:
fencedframe is the Privacy Sandbox replacement for iframe (Chrome 114+, HTML Living Standard). It is missing from the iframe-handling sites (`MANUALLY_CLEANED`, `SANITIZED_XPATH`, and `readability_lxml`'s early-drop list), so its content currently surfaces in extracted text.

Reproduction:

```python
html = '<html><body><article><p>real</p><fencedframe><p>ad</p></fencedframe><p>more</p></article></body></html>'
extract(html)  # 'real\nad\nmore' (with fix: 'real\nmore')
```

Adds fencedframe to those three sites. The test covers the cleaning path (fencedframe inside an article) and the readability fallback (fencedframe wrapping the article).

Found while looking at #720.


adbar:
The CI was failing but the main branch works again now, you'd have to merge it here in order to run the tests properly.

adbar:
LGTM, thanks.

PR #848, opened 2026-05-19
Improve title extraction

novucs:
- Always take title from head if it exists (and is not a domain)
- Only take first non-empty h1


adbar:
Please add a test to account for the changes you made.

novucs:
> Please add a test to account for the changes you made.

Thanks for the review. I've merged the latest changes & added tests

adbar:
Nice, thanks

PR #851, opened 2026-05-31
Fix inline <code> duplication in extracted text

### LINKED ISSUE #849: Inline <code> elements duplicated in extracted text
## Summary

When extracting text from HTML pages containing inline `<code>` elements inside paragraphs, the code content appears twice in the output — once from the parent element's text processing and once from the child `<code>` element being processed separately as a code block.

## Reproduction

Extract text from any page with inline code in paragraphs, e.g.:

```html
<p>Use <code>pip install trafilatura</code> to install.</p>
```

**Expected output:**
```
Use pip install trafilatura to install.
```

**Actual output:**
```
Use pip install trafilatura to install. pip install trafilatura
```

## Root Cause

Three separate code paths contribute:

1. **Main extraction pipeline** (`main_extractor.py`): `is_code_block_element()` returns `True` for all `<code>` elements including inline ones inside `<p>`, `<li>`, etc. This causes them to be extracted as standalone code blocks in addition to being included in the parent's text. The `_extract()` and `recover_wild_text()` functions also process inline `<code>` as separate sub-elements.

2. **Baseline fallback scraper** (`baseline.py`): When the main extractor falls back to baseline paragraph scraping, inline `<code>` text is included in the parent `<p>` text via `text_content()` but the `<code>` elements are not filtered from the tree first.

## Proposed Fix

- `is_code_block_element()`: early return `False` for `<code>` elements whose parent is a paragraph-like element (`p`, `li`, `td`, `th`, `span`, `a`)
- `_extract()` and `recover_wild_text()`: filter inline `<code>` from sub-element lists before passing to `handle_textelem`
- `baseline.py`: skip inline `<code>` elements when scraping text paragraphs

PR incoming.

bobtheskull-flameeyes:
### Problem

Inline `<code>` elements inside paragraphs (`<p>`, `<li>`, `<td>`, etc.) are extracted twice:

1. Once as part of their parent element's text flow
2. Once as a standalone code block

**Input:**
```html
<p>Use <code>pip install trafilatura</code> to install.</p>
```

**Expected output:**
```
Use pip install trafilatura to install.
```

**Actual output (before fix):**
```
Use pip install trafilatura to install. pip install trafilatura
```

### Root cause

`is_code_block_element()` returns `True` for **all** `<code>` tags regardless of context — it has no parent-context check. This causes inline code to be treated as standalone code blocks.

Additionally, the baseline fallback scraper (`baseline.py`) iterates `<code>` elements separately via `tree.iter()`, and `text_content()` on each produces text that's already captured by the parent `<p>`.

### Why the adjacent-element dedup (#881) doesn't fix this

PR #881 added a post-extraction dedup that drops elements whose text is identical to the immediately previous element. This **does not catch inline code duplication** because:

- The paragraph text (`"Use pip install trafilatura to install."`) != the code text (`"pip install trafilatura"`) — different strings, so exact-match dedup doesn't fire
- Short inline code like `<code>x</code>` is far below `MIN_DUPLICATE_LENGTH=50`
- The baseline scraper doesn't use the main extraction pipeline at all

### Fix (3 files, 1 commit)

**`trafilatura/main_extractor.py`:**
- `is_code_block_element()`: early-return `False` for `<code>` whose parent is a paragraph-like element (`p`, `li`, `td`, `th`, `dd`, `dt`)
- `_extract()` and `recover_wild_text()`: filter inline `<code>` from sub-element lists before processing

**`trafilatura/baseline.py`:**
- Skip inline `<code>` in the paragraph generator expression

### Tests

Comprehensive test coverage added to `tests/unit_tests.py`:

| Test | What it covers |
|------|---------------|
| `is_code_block_element` unit test | Returns `False` for inline contexts, `True` for block contexts |
| Single inline `<code>` in `<p>` | Basic duplication bug |
| Short inline `<code>` (`x`) | Proves dedup-only fix is insufficient (below `MIN_DUPLICATE_LENGTH`) |
| Multiple `<code>` in one `<p>` | `in`, `True`, `False` — none should leak as standalone blocks |
| `<code>` in `<li>` | List item context |
| `<code>` in `<td>` | Table cell context |
| Mixed inline + block code | Inline doesn't duplicate, block code still works |
| `fast=True` mode | Fast extraction path |
| `favor_recall=True` mode | Recovery path (`recover_wild_text`) |
| Baseline: standard case | `baseline()` scraper duplication |
| Baseline: short snippet | Single-char inline code |
| Baseline: multiple codes | All three keywords duplicated |

All 186 existing tests pass (0 regressions).

Fixes #849.

### DIFF
--- a/trafilatura/baseline.py
+++ b/trafilatura/baseline.py
@@ -76,6 +76,11 @@ def baseline(filecontent: Any) -> Tuple[_Element, str, int]:
     temp_text = ""
     # postbody = Element('body')
     for element in tree.iter('blockquote', 'code', 'p', 'pre', 'q', 'quote'):
+        # Skip inline <code> elements — their text is already captured by their parent <p>
+        if element.tag == 'code':
+            parent = element.getparent()
+            if parent is not None and str(parent.tag) in ('p', 'li', 'td', 'th', 'span', 'a'):
+                continue
         entry = trim(element.text_content())
         if entry not in results:
             SubElement(postbody, 'p').text = entry
--- a/trafilatura/main_extractor.py
+++ b/trafilatura/main_extractor.py
@@ -204,11 +204,15 @@ def handle_lists(element: _Element, options: Extractor) -> Optional[_Element]:
 
 def is_code_block_element(element: _Element) -> bool:
     "Check if it is a code element according to common structural markers."
+    # Inline <code> inside paragraph-like context is NOT a block element
+    parent = element.getparent()
+    if element.tag == "code" and parent is not None \
+       and parent.tag in ('p', 'li', 'td', 'th', 'span', 'a'):
+        return False
     # pip
     if element.get("lang") or element.tag == "code":
         return True
     # GitHub
-    parent = element.getparent()
     if parent is not None and "highlight" in parent.get("class", ""):
         return True
     # highlightjs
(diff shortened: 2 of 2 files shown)


adbar:
I didn't see your PR at the time. The code base has now evolved so here is what I suggest: see if the bug still exists, then fix the merge conflicts and merge the main branch (or start again if it's too messy).

adbar:
The PR is outdated and I believe this issue is now fixed.

PR #852, opened 2026-05-31
Fix comment content leaking into main text when include_comments=False

### LINKED ISSUE #850: Comment content leaks into main text when include_comments=False outside precision mode
## Summary

When `include_comments=False` (the default), comment sections from the HTML page still appear in the extracted main text unless `focus='precision'` is also set. Sites with substantial comment threads — such as LWN.net — are particularly affected, with entire comment trees appearing in the article body.

## Reproduction

```python
import trafilatura

# Fetch an LWN.net article with comments
html = trafilatura.fetch_url('https://lwn.net/Articles/...')
text = trafilatura.extract(html, include_comments=False)
# Comment content appears in `text`
```

## Root Cause

Two issues:

### 1. Comment nodes not pruned outside precision mode

In the core extraction flow, `prune_unwanted_nodes()` for comment XPaths is only called when `focus == 'precision'`. When `include_comments=False` but focus is not `'precision'` (i.e. the default), comment nodes remain in the tree and their text content is extracted as part of the main article.

### 2. `<details>` elements not matched by comment XPaths

Some sites (e.g. LWN.net) wrap comment threads in `<details class="CommentBox">` elements. The comment-related XPath expressions in `COMMENTS_XPATH` and `COMMENTS_DISCARD_XPATH` match `self::div`, `self::section`, and `self::list` but not `self::details`, so these comment blocks are never identified as comments in the first place.

## Proposed Fix

1. Move the `prune_unwanted_nodes` call for comments into the `else` branch (comments disabled) so comment nodes are always removed when `include_comments=False`, regardless of focus mode.

2. Add `self::details` to the 6 comment-related XPath expressions in `xpaths.py` so that `<details>` elements with comment-indicating class names are correctly identified.

PR incoming.

bobtheskull-flameeyes:
Fixes #850.

Comment sections leak into extracted main text when `include_comments=False` (the default) unless `focus='precision'` is also set.

### Changes

**`core.py`:**
- Move `prune_unwanted_nodes()` for comment XPaths into the `else` branch (comments disabled) so comment nodes are always removed when `include_comments=False`, regardless of focus mode

**`xpaths.py`:**
- Add `self::details` to the 6 comment-related XPath expressions (`COMMENTS_XPATH` and `COMMENTS_DISCARD_XPATH`) so that `<details>` elements with comment-indicating class names are correctly identified as comment blocks

### Tests

- All existing tests pass (no regressions)
- mypy clean (same error count as master)

### Motivation

Sites like LWN.net wrap comment threads in `<details class="CommentBox">` elements. Without `self::details` in the XPath expressions, these are never identified as comments. And even when comments are identified, they are only pruned in precision mode — not when `include_comments=False` is set with the default focus.

### DIFF
--- a/trafilatura/core.py
+++ b/trafilatura/core.py
@@ -280,6 +280,10 @@ def bare_extraction(
             )
         else:
             commentsbody, temp_comments, len_comments = Element("body"), "", 0
+            # When comments are disabled, prune comment nodes from the tree
+            # to prevent them from leaking into the main text extraction.
+            # Previously this only ran in "precision" mode.
+            cleaned_tree = prune_unwanted_nodes(cleaned_tree, REMOVE_COMMENTS_XPATH)
         if options.focus == "precision":
             cleaned_tree = prune_unwanted_nodes(cleaned_tree, REMOVE_COMMENTS_XPATH)
 
--- a/trafilatura/xpaths.py
+++ b/trafilatura/xpaths.py
@@ -64,23 +64,23 @@ BODY_XPATH = [XPath(x) for x in (
 
 
 COMMENTS_XPATH = [XPath(x) for x in (
-    """.//*[self::div or self::list or self::section][contains(@id|@class, 'commentlist')
+    """.//*[self::details or self::div or self::list or self::section][contains(@id|@class, 'commentlist')
     or contains(@class, 'comment-page') or
     contains(@id|@class, 'comment-list') or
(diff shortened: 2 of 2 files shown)


adbar:
Same as in your other PR: see if the bug still exists, then fix the merge conflicts and merge the main branch (or start again if it's too messy).

adbar:
superseded by #881

PR #864, opened 2026-06-13
fix: security hardening - certifi version pin and URL validation

yizhengjaredlim:
### Changes Made

1. **Pin certifi minimum version** (pyproject.toml)
   - Changed certifi to certifi >= 2024.0.0
   - Ensures up-to-date CA certificate bundle for HTTPS requests

2. **Add URL validation** (	rafilatura/downloads.py)
   - Added alidate_url() function that rejects private/internal IPs before making HTTP requests
   - Added validation call in etch_response() to prevent potential SSRF

3. **Remove global SSL warning suppression** (	rafilatura/downloads.py)
   - Moved urllib3.disable_warnings() from module-level scope

### Why

- Prevents SSRF attacks by rejecting requests to private IPs (127.0.0.1, 10.x.x.x, 192.168.x.x, etc.)
- Ensures proper TLS certificate validation with up-to-date CA bundle
- Removes dangerous global SSL warning suppression

All changes are backward-compatible.


adbar:
The tests won't pass because you use a local file in them.

I'd also suggest you to have a look at this function, the `courlan` package is already used to validate URLs:
https://github.com/adbar/courlan/blob/bb8ff2b267d80b7cc80da95d82609fe2ab3d70de/courlan/filters.py#L250

yizhengjaredlim:
Thanks for the feedback @adbar! I've updated the PR to use courlan.filters.is_valid_url instead of the custom validation function. The PR now uses the existing library's URL validation which is much cleaner.

adbar:
The general idea is sound but the PR does not meet the quality requirements for this repository.

PR #868, opened 2026-06-18
fix: keep Yoast FAQ block question headers

### LINKED ISSUE #471: Missing Yoast FAQ block headers
The paragraphs are there but not the question before (`div > strong class="schema-faq-section"`).

Example:
"Qui a inventé la robe portefeuille ?" missing from
https://web.archive.org/web/20240108165704/https://www.clicandfit.com/blog/conseils/robe-portefeuille/

cycsmail:
closes #471

Yoast FAQ blocks render each question as `<strong class="schema-faq-question">`. The bold question acts as a section header on the page, but trafilatura treated it as inline bold, so it got flattened or dropped depending on the options.

In `convert_tags` I retag those `<strong>` elements to `<head rend="h3">` so they survive extraction and come out as headers (### in Markdown, `<head rend="h3">` in XML). Added a test covering the default, formatted, and XML output.

### DIFF
--- a/trafilatura/htmlprocessing.py
+++ b/trafilatura/htmlprocessing.py
@@ -393,6 +393,12 @@ def convert_tags(tree: HtmlElement, options: Extractor, url: str | None = None) 
         for elem in tree.iter("a", "ref"):
             convert_link(elem, base_url)
 
+    # Yoast FAQ blocks: question headers are bold but act as titles (#471)
+    for elem in tree.xpath('.//strong[contains(@class, "schema-faq-question")]'):
+        elem.attrib.clear()
+        elem.set("rend", "h3")
+        elem.tag = "head"
+
     if options.formatting:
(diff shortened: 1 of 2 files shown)


trafilatura/htmlprocessing.py
```diff
@@ -393,6 +393,13 @@ def convert_tags(tree: HtmlElement, options: Extractor, url: str | None = None)
         for elem in tree.iter("a", "ref"):
             convert_link(elem, base_url)
 
+    # Yoast FAQ blocks: question headers are bold but act as titles (#471)
+    for elem in tree.iter("strong"):
```
adbar: This isn't particularly efficient, an XPath expression would be simpler and faster here.

cycsmail: Good call. Switched to `.//strong[contains(@class, "schema-faq-question")]` so it only walks the matching nodes instead of every `<strong>`. Updated the branch.

PR #873, opened 2026-06-26
docs: add a page explaining how to run the tests

### LINKED ISSUE #698: Docs: add page explaining how to run tests
So far part of the information is present in `CONTRIBUTING.md` and in the `tests/` folder.

It would be best to regroup all necessary instructions in a documentation page on how tests are run.

cycsmail:
Adds a `Running the tests` page to the docs, regrouping the install / pytest / mypy / ruff and benchmark instructions that were previously split between CONTRIBUTING.md and tests/README.rst, and wires it into the toctree. Also adds a small test guarding that the page exists and stays referenced in the tree.

The new docs test passes (2 passed) and ruff check/format are clean on the added file; mypy only targets the `trafilatura` package so the docs additions are out of its scope.

closes #698


adbar:
Good idea, it would be even better to add info about this in `CONTRIBUTING.md`. Not a link since it won't be valid as of now but a mention saying to check the docs?

cycsmail:
good idea, added a mention in CONTRIBUTING pointing at the testing docs. thanks!

adbar:
Thanks, the only remaining issue is that your current test doesn't add much functionality. Since we add tests for the docs I think it would be better to just test separately if they compile.

To avoid unnecessary checks I recommend this logic but it's open for discussion:
- `docs_tests.py` file in `docs/` folder with compilation test only (example [here](https://github.com/adbar/courlan/blob/84f9802613561e98238bbef0e2308c9c9965d15e/docs/test_docs.py))
- test in CI only for the Python version used for ReadTheDocs, example here:
https://github.com/adbar/courlan/blob/84f9802613561e98238bbef0e2308c9c9965d15e/.github/workflows/tests.yml#L62

cycsmail:
thanks, that makes more sense. switched to a sphinx compile test in docs/docs_tests.py (built with -W so broken refs fail) and wired it into CI only on 3.13 to match readthedocs. dropped the old string checks.

adbar:
Great, thanks!

PR #875, opened 2026-06-30
Quote metadata values to keep the Markdown header valid YAML (#814)

### LINKED ISSUE #814: Invalid YAML being produced by `--with-metadata` if there's a colon-space in the string
```
❯ trafilatura --with-metadata -u https://www.thehindu.com/sci-tech/energy-and-environment/cop30-a-beginners-guide-on-what-to-expect-from-the-climate-summit/article70259165.ece | head
---
title: COP30: a beginner’s guide on what to expect from the climate summit
author: Indu K Murthy
url: https://www.thehindu.com/sci-tech/energy-and-environment/cop30-a-beginners-guide-on-what-to-expect-from-the-climate-summit/article70259165.ece
hostname: thehindu.com
description: Explore what to expect from COP30, the critical climate summit addressing global commitments, finance, and adaptation strategies for climate action.
sitename: The Hindu
date: 2025-11-10
categories: ['Environment']
tags: ['climate change', 'climate change (politics)', 'United Nations (climate change)', 'global warming']
```

As `: ` indicates a key-value pair in YAML, it needs to be escaped.  The title should be in quotes: `title: "COP30: a beginner’s guide on what to expect from the climate summit"`.

gaoflow:
Fixes #814.

The Markdown/TXT metadata header interpolates each value raw:

```python
header += f"{attr}: {str(getattr(document, attr))}\n"
```

so a title like `COP30: a beginner's guide` is written as `title: COP30: a beginner's guide`, which isn't valid YAML (the `: ` reads as a second mapping key). Values are also silently reinterpreted on parse: `true`/`null` become a bool/null, a leading `#` starts a comment (→ null), a leading `&`/`*`/`[` is an anchor/alias/sequence, and so on.

`_yaml_scalar()` renders each scalar string safely: bare when it is unambiguous (ordinary titles, author names and URLs are unchanged), otherwise double-quoted via `json.dumps()`. A JSON string is always a valid YAML double-quoted scalar that round-trips exactly, so this stays dependency-free and covers colons, quotes, leading indicators, reserved words and numerics uniformly. `categories`/`tags` are lists and keep their existing `['a', 'b']` flow-sequence form, which is already valid YAML.

`test_markdown_metadata_yaml_safe` renders the scalars directly, round-trips a range of adversarial values through PyYAML (`importorskip`), and checks end-to-end that a `: `-containing title now produces a header that parses as valid YAML with the title preserved exactly. `pytest`, `mypy -p trafilatura` and `ruff check`/`ruff format` are clean.

Happy to add a HISTORY.md entry under whichever section you prefer.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -622,6 +622,49 @@ def test:
     assert "AP-fonden" in my_result
 
 
+def test_markdown_metadata_yaml_safe():
+    "Markdown metadata header must stay valid YAML for special values (GH #814)."
+    # scalar rendering: plain when safe, double-quoted (json) otherwise
+    assert core._yaml_scalar("Indu K Murthy") == "Indu K Murthy"
+    assert core._yaml_scalar("https://example.com/a:b") == "https://example.com/a:b"
+    assert core._yaml_scalar("élan vital") == "élan vital"
+    assert core._yaml_scalar("COP30: a guide") == '"COP30: a guide"'
+    assert core._yaml_scalar("#1 ranking") == '"#1 ranking"'
+    assert core._yaml_scalar("&launch") == '"&launch"'
+    assert core._yaml_scalar("true") == '"true"'
+    assert core._yaml_scalar("2024") == '"2024"'
+    assert core._yaml_scalar("[draft]") == '"[draft]"'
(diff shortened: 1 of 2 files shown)


adbar:
Thanks!

PR #877, opened 2026-07-04
Escape angle brackets in Markdown link destinations

gaoflow:
`_md_link` (`trafilatura/xml.py`) wraps a link/image target containing a space, `<`, `>`, `(` or `)` in a CommonMark angle-bracket destination `<...>`, but did not escape `<`/`>` inside it:

```python
safe = f"<{url}>" if any(c in url for c in " <>()") else url
```

Per CommonMark, an angle-bracket destination may not contain an unescaped `<` or `>`. So a URL like `http://x.com/a>b` had its target silently truncated at the `>` (href becomes `http://x.com/a`), and `http://x.com/a<b` voided the link entirely.

This backslash-escapes `\`, `<` and `>` (backslash first) inside the angle-bracket destination, following the same convention the file already uses for `~`, `|` and `[` in link text.

`test_markdown_link_angle_bracket_targets` round-trips the rendered Markdown back through a CommonMark parser (markdown-it-py) and asserts the decoded href equals the original, across URL classes with `>`, `<`, `<>` pairs, unbalanced parens, and the `None`/image branches. `ruff check` and `ruff format --check` are clean.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -2541,6 +2541,52 @@ def test_markdown_escaping():
     assert xml.xmltotxt(etree.fromstring(struck_img), True).strip() == "~~x ![A](i.jpg) y~~"
 
 
+def test_markdown_link_angle_bracket_targets():
+    "A '<' or '>' in a link/image target must stay inside the angle-bracket destination."
+    # each URL forces the angle-bracket form (space/paren/</>) and must round-trip:
+    # a bare '>' used to end the destination early (truncated URL), '<' voided the link
+    cases = {
+        "http://x.com/a>b": "[t](<http://x.com/a\\>b>)",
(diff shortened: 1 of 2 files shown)


adbar:
@gaoflow Good catch but how common are angle brackets in URLs? I don't think it really happens in the wild.

binggao1230:
Fair — in well-formed URLs it's rare, since `<`/`>` should be percent-encoded. The case I hit is trafilatura's actual input: arbitrary scraped HTML, where hrefs often aren't well-formed and can carry unencoded `<`/`>`. When they do, the current output is invalid Markdown (the link destination stops parsing at the `<`). The change only alters output when one of those chars is actually present, so it's a zero-cost safety net for normal URLs.

Happy to close if you'd still rather not carry it though — no strong feelings.

adbar:
Well, if you've seen the case in real life that's fine.

PR #878, opened 2026-07-05
docs: add example for adapting element lists in Python

### LINKED ISSUE #746: Documentation about settings could use examples
I'm trying to replace / modify things like `MANUALLY_CLEANED` and `MANUALLY_STRIPPED` to better suit my non-news-articles use case, but it's not really clear how or where to do that.  I'm sure I'll get there, but an example of getting at those either via modifying `DEFAULT_SETTINGS` before passing it into `extract()` and/or using a `settings.cfg` file would be helpful.

I'd prefer both examples, really, as I may need to change up some settings dynamically.

cycsmail:
closes #746

The settings docs note that some variables are Python lists but never show how to change them. Added a short section to docs/settings.rst with examples for modifying MANUALLY_CLEANED and MANUALLY_STRIPPED in place (append/remove/extend), plus a warning that reassigning the names has no effect since the lists are global to the process and the extraction functions read the existing objects.

Also added a test confirming that an in-place change to MANUALLY_CLEANED actually affects extraction output, and that removing the change restores the default.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -126,5 +126,32 @@ Here is how to change them:
 These remaining variables greatly alter the functioning of the package!
 
 
+Adapting element lists in Python
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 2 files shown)


HISTORY.md
```diff
@@ -40,6 +40,7 @@ Maintenance:
 - Tests: check consistency by @adbar (#859)
 - Modernize code base by @adbar (#860)
 - Format the code with ruff by @adbar (#861)
+- Document how to adapt element lists in Python (#746)
```
adbar: That's the wrong place to do it, the version 2.1.0 is already shipped. I'll update this file later.

adbar:
@cycsmail The PR looks good, please fix the point above.

cycsmail:
done, dropped the changelog line. thanks!

Issue #778, opened 2025-01-24
Deduplication is non-deterministic (and destructive)


BramVanroy:
In light of the large FineWeb-2 corpora, I came across an inconsistency in the data where extracted data seems not reproducible. After chasing it down, with help by @guipenedo, I found that this is caused by trafilatura's `deduplicate` option.

- When enabled, it does not seem to be deterministic: running `extract(... deduplicate=True)` on the same data can return different results
- Some (destructive) caching seems to be going on. In the code below, I have a loop that runs 11 times, and in each loop: a file's HTML contents are read, and trafilatura `extract` is called (with deduplication). But I find that sometimes after a few runs, the contents changes (often run 4 and 7). I find that in run 7, the extraction often ends up as `None`. 

So to me it seems that trafilatura  is doing some caching and after a while it will try to update the cache with new info, but in a destructive manner so that at each cache update, more data is removed. What I think is problematic is that this happens even with newly read data. As I said, at each iteration I read the file contents again from disk, and still some other types of caching seem to be happening that ignore the new input.

I am not entirely sure if this is expected behavior in the deduplication algorithm, but I do not think that many people would expect such behavior. No such issue occurs when deduplication is disabled.

Note: I am on `trafilatura==1.11.0` but I also tried on 2.0.0 and the issue persists.

---

Reproducible code (you can also decrease the number of items in `uid_and_file_paths`: https://gist.github.com/BramVanroy/c7e9778aa1f4259f7066e22e2cd1aa3a or here:

```python
from concurrent.futures import ProcessPoolExecutor, as_completed
from pathlib import Path
from tqdm import tqdm
import s3fs
from warcio.archiveiterator import ArchiveIterator
from trafilatura import extract
from commoncrawl_cc_annotation.utils import extract_uuid


crawl = 'CC-MAIN-2024-10'

uid_and_file_paths = [
    ['<urn:uuid:02111860-def3-4ebd-a7c6-1d414c5ff523>', 's3://commoncrawl/crawl-data/CC-MAIN-2024-10/segments/1707947474533.12/warc/CC-MAIN-20240224112548-20240224142548-00100.warc.gz'],
    ['<urn:uuid:0ba01026-a2f0-4edc-8af8-c70aad5ec4fa>', 's3://commoncrawl/crawl-data/CC-MAIN-2024-10/segments/1707948223038.94/warc/CC-MAIN-20240305060427-20240305090427-00899.warc.gz'],
[...]

PR #880, opened 2026-07-09
Fix small <article>/<main> bodies extracted twice (#879)

### LINKED ISSUE #879: Small <article>/<main> bodies extracted twice in 2.1.0 — minimal repro without <br> (follow-up to #768)
## Summary

`extract()` returns the entire body **twice** for small documents whose content sits in a semantic container (`<article>` or `<main>`). This looks like the same multiple-XPath-match lineage as #768 (closed 2026-06-18), but it still reproduces on 2.1.0 with a simpler trigger: no `<br>` or long-line requirement — just a small body (≤2 paragraphs) inside `<article>`/`<main>`.

## Environment

trafilatura 2.1.0 (latest PyPI), CPython 3.14, Linux.

## Minimal reproduction

```python

html = (
    "<html><body><nav>menu chrome</nav><article>"
    "<h1>The Example Chronicle</h1>"
    "<p>First synthetic paragraph of adequate length for extraction to engage properly.</p>"
    "<p>Second synthetic paragraph, also long enough to matter for the extractor.</p>"
    "</article><footer>footer chrome</footer></body></html>"
)
print(extract(html, output_format="markdown"))
```

Output: the heading once, then **both paragraphs twice**:

```
# The Example Chronicle

First synthetic paragraph of adequate length for extraction to engage properly.

Second synthetic paragraph, also long enough to matter for the extractor.

First synthetic paragraph of adequate length for extraction to engage properly.

Second synthetic paragraph, also long enough to matter for the extractor.
```

## Characterization

| Variant | Duplicated? |
| --- | --- |
| `<article>` wrapper, h1 + 2 `<p>` | yes |
| `<article>` wrapper, no heading, 2 `<p>` | yes |
| `<main>` wrapper, h1 + 2 `<p>` | yes |
| `<div>` wrapper, h1 + 2 `<p>` | **no** |
| `<article>` wrapper, h1 + 4 `<p>` | **no** |
| `<article>` with `<h2>` instead of `<h1>` | yes |

None of `fast=True`, `deduplicate=True`, `favor_precision=True` change the outcome. `output_format="txt"` behaves the same way.

So: semantic container + small body → doubled; generic container or a larger body → clean. Consistent with the container matching more than one extraction expression, as diagnosed in #768, with the small-body path still affected after that fix.

## Impact

Downstream consumers converting short articles/snippets get silently doubled content. We currently work around it with an exact-doubled-block-sequence collapse after `extract()`, which we'd gladly delete.

Happy to provide more variants if useful — the repro is fully deterministic.

apoorvdarshan:
Fixes #879.

## Root cause

For a small document whose main content sits in a semantic container (`<article>` or `<main>`), the main extractor collects the paragraphs into `result_body`. Because the body text is below `min_extracted_size`, `extract_content` then calls `recover_wild_text` on the backup tree, which re-appends the same `<p>` elements. The resulting body is `[head, p1, p2, p1, p2]`.

The existing post-recovery dedup loop only compared each block against the **immediately preceding** one (`current == previous`). Here the duplicates are **interleaved** (`p1, p2, p1, p2`), not adjacent, so the loop never matched and every paragraph was emitted twice.

A `<div>` wrapper or a larger body avoids this: a bare `<div>` isn't populated by the body XPaths (recovery then adds the paragraphs once), and a larger body clears `min_extracted_size` so recovery never runs.

## Reproduction

```python

html = (
    "<html><body><nav>menu chrome</nav><article>"
    "<h1>The Example Chronicle</h1>"
    "<p>First synthetic paragraph of adequate length for extraction to engage properly.</p>"
    "<p>Second synthetic paragraph, also long enough to matter for the extractor.</p>"
    "</article><footer>footer chrome</footer></body></html>"
)
print(extract(html, output_format="markdown"))
```

Wrong output on 2.1.0 / `master` (both paragraphs doubled):

```
# The Example Chronicle

First synthetic paragraph of adequate length for extraction to engage properly.

Second synthetic paragraph, also long enough to matter for the extractor.

First synthetic paragraph of adequate length for extraction to engage properly.

Second synthetic paragraph, also long enough to matter for the extractor.
```

Corrected output with this change:

```
# The Example Chronicle

First synthetic paragraph of adequate length for extraction to engage properly.

Second synthetic paragraph, also long enough to matter for the extractor.
```

`<main>` reproduces identically and is likewise fixed.

## Fix

In `extract_content`, the post-recovery dedup now compares each substantial block against **all earlier** blocks (a `seen` set) instead of only the immediately preceding one. It keeps the existing `DUPLICATE_ADJACENT_LIMIT` length gate, so short genuine repeats still survive (per #778).

This is deliberately localized to the dedup loop rather than to `recover_wild_text`: the loop runs **after** `temp_text` length is captured for the baseline-fallback decision, so removing the visual duplicate does not push the (correct) text below `min_extracted_size` and does not alter which downstream path runs — formatting and the fallback threshold are preserved. Fixing `recover_wild_text` directly instead would drop the deduplicated text under the threshold and trigger the lossy baseline path, degrading the `<article>` output.

## Tests

- Extended `test_no_duplicate_content` in `tests/unit_tests.py` with the `<article>` and `<main>` triggers, asserting each paragraph appears exactly once. It fails on unmodified `master` (`assert 2 == 1`) and passes with this change.
- Verified short genuine repeats (below the length gate) and distinct multi-paragraph content are unaffected.
- `pytest tests/unit_tests.py tests/baseline_tests.py tests/deduplication_tests.py tests/filters_tests.py tests/metadata_tests.py tests/realworld_tests.py tests/json_metadata_tests.py tests/xml_tei_tests.py` — 186 passed, 2 skipped.
- `ruff check .` clean, `ruff format --check` clean, `mypy trafilatura/` introduces no new errors (the single pre-existing `zstandard` stub error is unrelated).

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -1957,6 +1957,23 @@ def test_no_duplicate_content():
     assert (extract(dup768, output_format="txt", config=real_config) or "").count("Line that has to have") == 1
     dup817 = "<html><body><div id='content'><p>Authoritative taxonomy of but let us leave it as it is 1 2 3</p></div><p>some text long enough not to skip and printed twice on this line some text long enough not to skip and printed twice on this line</p></body></html>"
     assert (extract(dup817, output_format="txt", config=real_config) or "").count("Authoritative taxonomy") == 1
+    # #879: a small <article>/<main> body (below min_extracted_size) triggers wild-text
+    # recovery which must not re-append paragraphs already collected by the main extractor
+    dup879 = (
+        "<html><body><nav>menu chrome</nav><article>"
+        "<h1>The Example Chronicle</h1>"
+        "<p>First synthetic paragraph of adequate length for extraction to engage properly.</p>"
+        "<p>Second synthetic paragraph, also long enough to matter for the extractor.</p>"
+        "</article><footer>footer chrome</footer></body></html>"
+    )
+    out879 = extract(dup879, output_format="txt", config=real_config) or ""
+    assert out879.count("First synthetic paragraph") == 1
+    assert out879.count("Second synthetic paragraph") == 1
+    # same trigger with a <main> wrapper
+    dup879_main = dup879.replace("article>", "main>")
+    out879_main = extract(dup879_main, output_format="txt", config=real_config) or ""
+    assert out879_main.count("First synthetic paragraph") == 1
+    assert out879_main.count("Second synthetic paragraph") == 1
 
 
 def test_list_item_block_child_single_bullet():
--- a/trafilatura/main_extractor.py
+++ b/trafilatura/main_extractor.py
@@ -774,15 +774,18 @@ def extract_content(cleaned_tree: HtmlElement, options: Extractor) -> tuple[_Ele
     if len(result_body) == 0 or len(temp_text) < options.min_extracted_size:
         result_body = recover_wild_text(backup_tree, result_body, options, potential_tags)
         temp_text = " ".join(result_body.itertext()).strip()
-    # drop substantial elements repeating the previous one (overlapping-candidate / recovery artifact);
-    # length-gated so short genuine repeats stay for the dedup (#778) and tree-size guards
-    previous = None
+    # drop substantial elements repeating an earlier one (overlapping-candidate / recovery artifact);
+    # length-gated so short genuine repeats stay for the dedup (#778) and tree-size guards.
+    # #879: wild-text recovery can re-append blocks already collected by the main extractor,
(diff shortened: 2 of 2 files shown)


adbar:
Thanks for the PR, I'm currently working on the same problem and I'm not sure this here is the best approach. I'll leave your PR parked until I know more.

adbar:
Superseded by #881, sorry for that.

PR #891, opened 2026-07-25
fix: keep sup and sub boundaries in Markdown output

### LINKED ISSUE #889: Markdown output: <sup>/<sub> text concatenates with no marker (100²=10000 becomes 1002=10000)
### Summary
Superscript and subscript content is emitted directly adjacent to the surrounding text, with no marker, so `100<sup>2</sup>=10000` becomes the unreadable run-on `1002=10000` and a dated annotation `2011<sub>15ya</sub>` becomes `201115ya`. The distinction between base and exponent/index is lost, and the numbers are silently wrong.

### Minimal repro
```python
import trafilatura

html = """<!DOCTYPE html><html><head><title>Exponents</title></head><body>
<article>
<p>The deepest layer has 100<sup>2</sup>=10000 terminal nodes.</p>
<p>Written 2011<sub>15ya</sub> during a long winter.</p>
</article></body></html>"""

print(trafilatura.extract(html, output_format="markdown", favor_recall=True))
```

### Observed
```
The deepest layer has 1002=10000 terminal nodes.

Written 201115ya during a long winter.
```

### Expected
Some separator that preserves the sup/sub boundary, e.g. `100^{2}=10000` and `2011_{15ya}` (or `100^2`, `2011~15ya~`, etc.). Any marker is better than none; the current output changes the numeric meaning.

### Root cause / pointer
`trafilatura/htmlprocessing.py:36-37` maps `sub` → `#sub` and `sup` → `#sup` as `rend` values on `hi` elements. The Markdown serializer in `trafilatura/xml.py` only wraps a `hi` element's text when its `rend` is a key of `HI_FORMATTING`:

- `trafilatura/xml.py:80` — `HI_FORMATTING = {"#b": "**", "#i": "*", "#u": "__", "#t": "`"}` (no `#sup`/`#sub`)
- `trafilatura/xml.py:304-305` — `if rend in HI_FORMATTING: elem_text = f"{HI_FORMATTING[rend]}{elem_text}{HI_FORMATTING[rend]}"`

Because `#sup`/`#sub` are absent from `HI_FORMATTING`, the branch is skipped and the text is concatenated markerless. This is unconditional — it fires on any page with sup/sub.

### Suggested fix
Add asymmetric-marker handling for `#sup`/`#sub` in the Markdown path (they need distinct open/close markers, unlike the symmetric entries already in `HI_FORMATTING`), e.g. emit `^{...}` / `_{...}` or `<sup>`/`<sub>` passthrough.

## Reproduction environment
- trafilatura 2.1.0, lxml (bundled), Python 3.13/3.14
- Observed against real archived pages; the minimal HTML above reproduces it in isolation.

igorsaevets:
Fixes #889.

- Emit `sup`/`sub` as inline HTML in Markdown so exponent and index boundaries survive:
  `100<sup>2</sup>=10000` was serialized as `1002=10000`, silently changing the number.
  `HI_FORMATTING` only holds symmetric markers, so these two never entered the branch.
- Drop an empty `<sup></sup>` during tag conversion while preserving its tail text.
  `process_node()` hands a textless element the following text as its own, so without this
  the new marker would wrap the wrong words (`a<sup></sup>b` became `a<sup>b</sup>`).
  Doing it in `convert_tags()` rather than via `CUT_EMPTY_ELEMS` keeps the tail in every
  focus mode, including `favor_precision`.
- Add regression coverage for normal, empty, whitespace-only and precision-mode inputs.

I kept inline HTML rather than the pandoc-style `^2^` / `~15ya~`: CommonMark passes raw
inline tags through, and GitHub renders a single tilde as strikethrough, so `~15ya~` would
have replaced one wrong meaning with another.

Note on reproducing: the short sample in the issue falls below `MIN_EXTRACTED_SIZE` and
enters the fallback path, which prints `100\n2\n=10000`. The regression tests use the
zero-threshold config to exercise the Markdown serializer directly; with content above the
threshold, master produces exactly the concatenation reported in #889.

Validation on this branch: `pytest` (239 passed, 3 skipped, including `realworld_tests.py`),
`ruff check .` and `mypy -p trafilatura` unchanged from master, `ruff format --check` clean.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -721,6 +721,35 @@ def test_markdown_list_item_inline_spacing():
     assert extract(htmlstring, output_format="markdown", config=ZERO_CONFIG) == "1. Foo *bar* baz."
 
 
+def test_markdown_sup_sub_keep_boundary():
+    """sup/sub must keep a boundary, else 100<sup>2</sup> reads as 1002 (issue #889)"""
+    sup = "<html><body><article><p>The layer has 100<sup>2</sup>=10000 nodes.</p></article></body></html>"
+    sub = "<html><body><article><p>Written 2011<sub>15ya</sub> in winter.</p></article></body></html>"
+    assert extract(sup, output_format="markdown", config=ZERO_CONFIG) == "The layer has 100<sup>2</sup>=10000 nodes."
+    assert extract(sub, output_format="markdown", config=ZERO_CONFIG) == "Written 2011<sub>15ya</sub> in winter."
+    # flanking whitespace stays outside the tags, exactly as for the symmetric markers
+    spaced = "<html><body><article><p>x <sup> 2 </sup> y</p></article></body></html>"
+    bold = "<html><body><article><p>x <b> 2 </b> y</p></article></body></html>"
+    assert extract(spaced, output_format="markdown", config=ZERO_CONFIG) == "x  <sup>2</sup>  y"
+    assert extract(bold, output_format="markdown", config=ZERO_CONFIG) == "x  **2**  y"
+
+
+def test_markdown_empty_sup_sub_are_dropped():
+    """An empty sup/sub must be removed without taking the text after it (issue #889)"""
+    # process_node() hands a textless element its tail as text, so without the removal in
+    # convert_tags() the new marker would wrap the *following* words: a<sup>b</sup>
+    for tag in ("sup", "sub"):
+        doc = f"<html><body><article><p>a<{tag}></{tag}>b</p></article></body></html>"
+        assert extract(doc, output_format="markdown", config=ZERO_CONFIG) == "ab"
+    # a footnote marker whose only child is dropped must still not swallow the sentence
+    footnote = '<html><body><article><p>Fact<sup><img src="x.png"/></sup> follows here.</p></article></body></html>'
+    assert extract(footnote, output_format="markdown", config=ZERO_CONFIG) == "Fact follows here."
+    # the tail survives in precision mode too, where CUT_EMPTY_ELEMS pruning would drop it
+    tailed = "<html><body><article><p>alpha<sup></sup>beta gamma.</p></article></body></html>"
+    assert extract(tailed, output_format="markdown", config=ZERO_CONFIG) == "alphabeta gamma."
+    assert extract(tailed, output_format="markdown", favor_precision=True, config=ZERO_CONFIG) == "alphabeta gamma."
+
+
 def test_extract_with_metadata():
     """Test extract_with_metadata method"""
     url = "http://aa.bb/cc.html"
--- a/trafilatura/htmlprocessing.py
+++ b/trafilatura/htmlprocessing.py
@@ -409,6 +409,13 @@ def convert_tags(tree: HtmlElement, options: Extractor, url: str | None = None) 
         elem.set("rend", "h3")
         elem.tag = "head"
 
+    # an empty sup/sub carries nothing to raise or lower, and process_node() would later
+    # hand it the following text as its own, so the marker would wrap the wrong words.
+    # The tail is kept in every focus mode, unlike the pruning of CUT_EMPTY_ELEMS.
+    for elem in tree.iter("sub", "sup"):
+        if not elem.text and len(elem) == 0:
+            delete_element(elem)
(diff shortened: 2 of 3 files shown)


adbar:
@igorsaevets Your PR looks interesting but please make sure the CI passes.

igorsaevets:
Thanks for looking. I dug into the failures and they do not come from this branch.

**The lint step is failing on unmodified `master`.** [ruff 0.16.0](https://astral.sh/blog/ruff-v0.16.0), released 2026-07-23, widened its default rule selection from 59 rules to 413, and `pyproject.toml` asks for an unpinned `ruff >= 0.15` with no `[tool.ruff.lint] select`. Checked out 3d89493 with nothing else applied:

```
ruff 0.15.4   ruff check .   ->  All checks passed!
ruff 0.16.0   ruff check .   ->  Found 103 errors.
```

The 103 findings are in 31 files across `BLE001`, `I001`, `FURB167`, `UP031`, `SIM102` and others. This branch only touches `htmlprocessing.py`, `xml.py` and `tests/unit_tests.py`, and contributes none of them.

**The tests themselves pass.** The `windows-latest` job on this PR ran the full suite: `289 passed, 5 skipped, 9 warnings in 98.97s`.

**Why only Windows was green, which is worth a look on its own.** That job installed the same `ruff==0.16.0` and its log also ends with `Found 103 errors.` — but the step still reported success. The lint step runs two commands in one `run:` block, and on Windows that block is executed by PowerShell, where the exit status of the step is taken from the last command only. `ruff format --check` passes, so the `ruff check` failure is discarded. On Linux and macOS the block runs under `bash -e` and aborts on the first failure. So lint has effectively not been enforced on the Windows leg.

I opened #892 with the one-line-ish fix for the first part: select `E4`, `E7`, `E9`, `F` explicitly in `pyproject.toml`, which is what ruff enforced before 0.16 and does not move when defaults change again. Same reasoning as the note already in that file about pinning mypy options individually. If you would rather cap the tool with `ruff >= 0.15, < 0.17`, or adopt some of the new rules deliberately, say which and I will redo it that way.

Once one of those lands I will rebase this branch so the checks run clean, and it can be judged on the actual diff.

igorsaevets:
Following up with evidence instead of more argument.

I ran your matrix on my own fork, so you can see the result without having to approve a workflow run for a first-time contributor. Both branches are built from `3d89493` with nothing else applied:

| branch | `Lint and format check` | `Test with pytest` |
|---|---|---|
| master + the `pyproject.toml` change from #892 | `All checks passed!` | `291 passed, 1 skipped` |
| the same, plus this PR's commit | `All checks passed!` | `293 passed, 1 skipped` |

- https://github.com/igorsaevets/trafilatura/actions/runs/30354769724 (#892 alone)
- https://github.com/igorsaevets/trafilatura/actions/runs/30354767272 (#892 + this PR)

The two extra tests are the ones this PR adds. In the second run, `Lint and format check`, `Type checking`, `Test with pytest` and `Test docs` are all green on every job. The one red step is `Upload coverage to Codecov`, failing with `Token required - not valid tokenless upload` - a fork-only condition, since codecov has no token for my copy. It does not arise here.

So the honest answer to "make sure the CI passes" is that this branch cannot go green on its own: the lint step aborts on 103 findings in 31 files it does not touch, before pytest ever runs. #892 is what unblocks it.

I have not touched this branch and will not rebase it until you have decided how you want the ruff selection handled - pin the rule set, cap the version, or adopt some of the new rules deliberately. Happy to redo #892 whichever way you prefer, or to close it if you would rather fix it your own way.

adbar:
Yes, that's correct, you'd now need to merge main here.

igorsaevets:
Done, `master` merged in (`7fd92e7`). All 11 checks are green, including `build (windows-latest, 3.11)` and CodeQL.

I ran your matrix locally against `ruff 0.16.0` before pushing, since that is what the `dev` extra resolves to now:

```
ruff check .                                  ->  All checks passed!
ruff format --check --diff trafilatura tests  ->  41 files already formatted
pytest                                        ->  289 passed, 5 skipped
```

The diff itself is unchanged: `trafilatura/htmlprocessing.py`, `trafilatura/xml.py`, and two tests in `tests/unit_tests.py`.

PR #904, opened 2026-08-13
Fix 2.2.0 regression dropping requested images

vvsotnikov:
This fixes a regression in Trafilatura 2.2.0. Short documents containing
requested images can be replaced by equal or shorter text-only fallbacks,
which removes `<graphic>` elements from XML output despite
`include_images=True`.

## Reproduction

```python
import trafilatura

html = """
<html><body><article>
  <p>Some reasonably long paragraph of body text that trafilatura will keep because it
  is clearly the main content of this document and not navigation furniture.</p>
  <img src="https://cdn.test/a.png" alt="a chart">
  <p>A second paragraph that follows the figure and continues the argument at length.</p>
</article></body></html>
"""

print(trafilatura.extract(
    html,
    output_format="xml",
    include_images=True,
    include_links=True,
))
```

Run the script against both releases:

```console
uv run --isolated --no-project --with trafilatura==2.1.0 python repro.py
uv run --isolated --no-project --with trafilatura==2.2.0 python repro.py
```

Observed results:

- 2.1.0 includes `<graphic src="https://cdn.test/a.png" alt="a chart"/>`.
  It also exhibits the older duplicate-text behavior.
- 2.2.0 returns one text-only paragraph and omits `<graphic>`.
- This PR preserves `<graphic>` without restoring the duplicated text.

The extracted text is shorter than the default `MIN_EXTRACTED_SIZE`. Justext
and then the baseline extractor can replace the structured result even when
they do not recover more text.

This change keeps an existing image when a text fallback does not recover more
text. A longer fallback still wins, so genuine content recovery is unchanged.

## Verification

- `pytest -q`: 335 passed, 1 skipped
- `ruff check .`: passed
- `ruff format --check trafilatura tests`: passed
- `python tests/eval_gate.py`: passed
  - `fast`: F1 0.9184
  - `fallback`: F1 0.9243

Disclosure: prepared with AI assistance and reviewed and verified locally.

### DIFF
--- a/tests/unit_tests.py
+++ b/tests/unit_tests.py
@@ -3,6 +3,7 @@
 Unit tests for the trafilatura library.
 """
 
+import json
 import logging
 import sys
 import time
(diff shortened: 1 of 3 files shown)


adbar:
Thanks!

PR #911, opened 2026-08-21
fix: drop infinite-scroll articles appended below the real one

cvl01:
## Problem

News themes with infinite scroll append the *next* 2-3 full articles into a container below the real one. Their text was extracted as part of the current article.

Reproducer: https://kulu-media.com/meta-outage-hits-facebook-instagram-and-messenger/ (Zox News theme, `#mvp-post-add-box`) returns the article plus three unrelated ones.

## Why not `OVERALL_DISCARD_XPATH`

Adding a token there does not stick, for two reasons:

1. On these pages the appended articles are most of the text (1.7k real chars out of 11.8k in the reproducer), so `prune_unwanted_nodes()`'s `new_len > old_len / 7` guard treats a correct prune as over-pruning and restores the backup. Loosening that ratio cost ~0.002 F1 on the `fast` benchmark, so it is left alone.
2. Even with the guard satisfied, `compare_extraction()` only prunes `raw_tree` in precision mode, so readability still saw all four articles and picked the longest appended one over the real one.

## Fix

A dedicated `REMOVE_APPENDED_ARTICLES_XPATH`, pruned on the raw tree in `trafilatura_sequence()` following the existing `REMOVE_COMMENTS_XPATH` pattern. Unguarded, and every stage inherits it (main extractor, readability, justext, baseline). Matches `mvp-post-add-(box|wrap)` and `infinite-?scroll` on `@id` or `@class`.

## Verification

- Fixed in all four modes (default, fast, recall, precision).
- Own-benchmark F1 improves on both runners: `fast` 0.9184 -> 0.9195, `fallback` 0.9243 -> 0.9253. The floors in `eval_baseline.json` are left untouched; happy to re-pin them upward if you prefer.
- Full test suite passes (341). ruff and mypy clean.
- Regression test added in `realworld_tests.py` with a trimmed mock page. Verified non-vacuous: it fails on all 3 parametrizations without the `core.py` change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adbar:
Nice, thanks!

PR #915, opened 2026-08-25
fix: escape literal CommonMark syntax characters in output (#914)

### LINKED ISSUE #914: Markdown output: literal CommonMark syntax characters in extracted text are misrendered (asterisks, brackets, headings, etc.)
## Summary

`xmltotxt()` (markdown mode, `include_formatting=True`) emits CommonMark-meaningful characters — `*`, `_`, `` ` ``, `~`, `[`, `]`, `<` — verbatim whenever they occur in extracted prose, instead of escaping them. Any renderer consuming trafilatura's markdown output misinterprets these as emphasis, code spans, link/image brackets, or raw HTML/autolinks instead of literal characters from the source page.

Separately, a paragraph, list item, or blockquote whose text happens to *start* with `#`, `>`, `-`, `+`, or a digit run followed by `.`/`)` is misread as a heading, blockquote, or list marker, because nothing distinguishes "real" block syntax (added by trafilatura itself) from a line that merely starts with the same character by coincidence.

This affects any consumer that renders trafilatura's `markdown` output format through a CommonMark parser (docs pipelines, chat/RAG ingestion, static site generators, etc.).

## Example failures

**1. Inline asterisks read as emphasis**
Source HTML: `This *should* not be italic and 3*4=12`
Markdown emitted: `This *should* not be italic and 3*4=12`
Rendered HTML: "This <em>should</em> not be italic and 3<em>4=12"</em> — the literal asterisks from the source page are silently turned into italics and an unclosed emphasis run.

**2. Leading character read as block syntax**
Source HTML: `# not a heading` (e.g. a tweet or comment quoting a hashtag-like string)
Markdown emitted: `# not a heading`
Rendered HTML: `<h1>not a heading</h1>` instead of a plain paragraph. The same failure mode applies to text starting with `>` (turned into a blockquote), `-`/`+` (turned into a bullet list), or `1.`/`1)` (turned into an ordered list).

**3. Brackets/backticks/underscores corrupt inline text and links**
Source HTML: "a_b_c and \`code\` and [bracket] and <tag> text"
Markdown emitted: "a_b_c and \`code\` and [bracket] and <tag> text"
Rendered HTML: "a<em>b</em>c and <code>code</code> and " with `[bracket]` left as a dangling reference-style link shell and `<tag>` interpreted as raw/unknown HTML — none of which existed in the original source text.

## Expected behavior

Literal occurrences of these characters in extracted prose should be backslash-escaped in the markdown output so a CommonMark renderer reproduces them as plain text, while trafilatura's own emphasis/strikethrough/link/heading/list markup continues to render as intended.

scottwulf:
Stray CommonMark-meaningful characters in extracted prose - *, _, `, ~, [, ], < - were emitted verbatim, causing renderers to misread them as emphasis, code spans, link/image brackets, or raw HTML/autolinks. A leading #, >, -, +, or digit+.")" at the start of a paragraph, list item, or blockquote line was likewise misread as heading, blockquote, or list syntax. Update the one real-world fixture whose expected output contained a literal asterisk.

Escape these via a tree-mutation pass (_escape_inline_specials_tree) that runs before markdown markup is generated, so real **bold**/*italic*/~~strikethrough~~ syntax and link/image brackets are never touched. Verbatim \<code> and inline-code (#t) spans are skipped since their content must stay literal, and plain-text mode (include_formatting=False) is unaffected. A separate _escape_block_start pass handles line-leading heading/blockquote/list markers for p/quote/item elements only, since table cells are inline-only in GFM. _md_link now escapes only not-yet-escaped brackets so it stays safe whether or not the tree-wide pass already ran. The del/~~ escaping now relies on the tree-wide pass instead of its own ad hoc replace.

Add test_markdown_asterisk_escaping and test_markdown_special_char_escaping covering paragraphs, headings, list items, table cells, links, images, verbatim code, nested formatting, backslash round-tripping, and block-start marker escaping.

Closes #914


adbar:
Hi @scottwulf, your PR looks good but there is a minor issue with code linting. Could you please look into it?

scottwulf:
Sorry @adbar. I missed staging two lint fixes, the one you referenced on line 3470 and another on line 3441. Fixed.

P.S. Also sorry for the force-push on this multi contributor repo.

adbar:
Thanks!

PR #917, opened 2026-08-25
perf: skip full compression detection when loading local files

### LINKED ISSUE #595: `utils.decode_file()`: add switch for full detection or GZip only
The use cases differ, it is not always necessary to test a binary string against all possible compression methods:

1. Use in `downloads` and `spider` modules: Run full detection
2. Use in `utils.load_html()`: Just test if it's a GZip file (i.e. downloaded and archived by Trafilatura)

ChrisJr404:
Closes #595.

`decode_file()` always runs the full gzip/zstd/brotli/zlib cascade, but `load_html()` only ever gets local input, where the only compression in practice is a GZip file archived by Trafilatura. So this adds a `full_detection` flag (default `True`, so `downloads` and `spider` keep the full cascade) and `load_html()` now passes `full_detection=False` to test for GZip only.

Extended `test_decode` to cover the GZip-only path.


adbar:
Thanks, I'm not sure this PR is beneficial: `load_html()` is also the entry point for `extract()`, so users can pass arbitrary compressed bytes (zstd, brotli) to it. This could silently break.
The existing cascade already uses magic-number checks (a few byte comparisons), so the speed gain would be fairly small compared to what we stand to lose.

ChrisJr404:
Good point, I had not accounted for `load_html()` being the shared entry point for `extract()`. Skipping the full compression detection there would risk silently mishandling zstd/brotli input that users legitimately pass in, and since the existing cascade is only a few magic-number comparisons the speed win is small next to that regression risk. Not worth the tradeoff. Closing this; thanks for the careful review.

PR #920, opened 2026-08-28
fix: decompress zstd frames without a declared content size

### LINKED ISSUE #912: zstd responses without a declared content size are silently returned undecompressed
### Summary

When `zstandard` is importable, trafilatura advertises `zstd` in `Accept-Encoding`. If the
server then returns a zstd frame **without a declared content size** — which is what
streaming/dynamic origins produce — `fetch_url()` returns the still-compressed bytes,
lossily decoded to a `str`. `extract()` and `extract_metadata()` then return `None`, with
no error raised and nothing logged above `WARNING`.

The failure is silent: you get a plausible-looking non-empty `str` back, so a caller has no
signal that anything went wrong short of noticing the extraction is empty.

### Environment

- trafilatura 2.2.0 (relevant code identical on current `master`)
- Python 3.14.7, Linux x86_64
- zstandard 0.25.0, urllib3 2.7.0, pycurl 7.45.7 / libcurl 8.18.0

### Reproduction

```python
import trafilatura

dl = trafilatura.fetch_url("https://interconnected.org/home/2023/03/16/singularity")
print(dl.encode("utf-8", "surrogateescape")[:4])  # b'(\xb5/\xfd'  -> zstd magic
print(trafilatura.extract(dl))                    # None
```

Reproduces identically on `https://simonwillison.net/`, `https://waxy.org/` and
`https://kottke.org/`. On all four, `zstandard.get_frame_parameters(body).content_size`
is `CONTENTSIZE_UNKNOWN`.

Expected: the decompressed HTML, and article text from `extract()`. Fetching the same URL
with `urllib.request` + `Accept-Encoding: gzip` and decompressing explicitly yields 11,152
characters of correctly extracted text, so neither the page nor the extractor is at fault.

### Root cause

The decisive bug is in `handle_compressed_file()` (`trafilatura/utils.py`, zstd branch):

```python
if HAS_ZSTD and filecontent[:4] == b"\x28\xb5\x2f\xfd":
    try:
        return zstandard.decompress(filecontent)  # max_output_size=???
    except zstandard.ZstdError:
        LOGGER.warning("invalid ZSTD file")
```

One-shot `zstandard.decompress()` requires the frame header to declare the decompressed
size. Streamed responses omit it, so this raises

```
ZstdError: could not determine content size in frame header
```

which is caught, logged, and then falls through — returning `filecontent` **unchanged**.
The caller cannot distinguish "decompressed successfully" from "gave up", and `decode_file()`
proceeds to string-decode the compressed bytes.

Verified directly:

```python
>>> handle_compressed_file(raw)  # 11451 zstd bytes in
                                 # 11451 identical bytes out, no exception
```

Both downloaders reach this path, for separate reasons, so it is not specific to either:

- **pycurl** (`_send_pycurl_request`, `downloads.py`): `Accept-Encoding` is set by hand via
  `CURLOPT_HTTPHEADER`, but `curl.setopt(pycurl.ACCEPT_ENCODING, ...)` appears only as a
  comment. Without `CURLOPT_ACCEPT_ENCODING`, libcurl does not decompress what it asked for,
  so `perform_rb()` hands back raw bytes.
- **urllib3** (`_send_urllib_request`): `Response.data` is still zstd for the same URL
  (verified — magic bytes `28b52ffd`). This looks like a urllib3-side issue with
  `read_chunked()` not normalising `decode_content=None` the way `read()` does, so it is
  probably not yours to fix — but it does mean *removing pycurl is not a workaround*.

Both then converge on `decode_file()` → `handle_compressed_file()`, where the silent
fall-through turns a recoverable situation into an empty extraction.

### Suggested fix

Use a decompressor that does not require a declared size:

```python
return zstandard.ZstdDecompressor().stream_reader(filecontent).read()
```

Verified locally against all four failing responses. This alone repairs both downloader
paths, since it removes the shared silent failure.

Two smaller points, at your discretion:

- Consider enabling `curl.setopt(pycurl.ACCEPT_ENCODING, "")` so libcurl decompresses what
  the request advertised. (Note libcurl is frequently built without zstd — mine is — so the
  manual path still has to work regardless.)
- Consider making the fall-through non-silent. Returning the input unchanged on a
  decompression failure is indistinguishable from success at every call site.

### Why this may have gone unreported

`zstandard` is not a default dependency — it is in the `all` extra — and trafilatura only
advertises `zstd` when it is importable (`downloads.py`: `if HAS_ZSTD and "zstd" not in
DEFAULT_HEADERS["accept-encoding"]`). So a plain `pip install trafilatura` never requests
zstd and never hits this.

It surfaces for `trafilatura[all]` users, and for anyone whose environment has `zstandard`
for unrelated reasons — in my case Fedora's `rpmlint` depends on it unconditionally, so a
plain trafilatura install picked it up from the system site-packages.

audi0417:
Fixes #912.

### The bug

`handle_compressed_file()` used the one-shot `zstandard.decompress()`, which
requires the frame header to declare the decompressed size. Servers that
compress responses on the fly omit it, so the call raised `ZstdError`, the
warning was logged, and the function fell through to returning the payload
unchanged. `decode_file()` then string-decoded the compressed bytes, so a
caller received a plausible-looking non-empty `str` while `extract()` returned
`None` — with nothing logged above `WARNING` to signal it.

Checked against the four sites named in the issue. On `master` all four come
back as the zstd magic lossily decoded to a `str` and extract to nothing;
with this change they extract 18421 / 25598 / 4732 / 9178 characters.

### The fix

Decompress through a streaming decompressor, which does not need the size up
front, and iterate over `unused_data` so concatenated frames are handled as
well.

A payload whose last frame never reaches `eof` is rejected. This part matters:
the obvious alternative, `stream_reader(..., read_across_frames=True).read()`,
returns a *partial* document for a truncated frame rather than raising — a
half-received response would have been silently accepted as complete, trading
one silent failure for a worse one. Measured on a truncated multi-block
payload it returns 786 KB of a 1.7 MB document with no error.

### Tests

The existing case only covered `zstandard.compress()`, which does declare the
size — hence the gap. Added coverage for a streamed frame, concatenated
frames, and a truncated frame that must not be accepted. The new assertions
fail on `master` and pass with this change.

`pytest` (343 passed), `mypy -p trafilatura`, `ruff check`, `ruff format
--check` and `tests/eval_gate.py` (F1 unchanged at the pinned floor) are all
clean.

### Note on the downloaders

I left both downloaders alone. `handle_compressed_file()` is the common
chokepoint — it deliberately does not trust response headers — so fixing it
covers the pycurl and urllib3 paths at once. The commented-out
`CURLOPT_ACCEPT_ENCODING` in `_send_pycurl_request` looks like a deliberate
choice and is a separate performance question, not a correctness one.

### DIFF
--- a/tests/downloads_tests.py
+++ b/tests/downloads_tests.py
@@ -293,6 +293,12 @@ def test_config():
     assert custom["Cookie"] == "yummy_cookie=choco; tasty_cookie=strawberry"
 
 
+def zstd_stream_compress(data: bytes) -> bytes:
+    "Compress data into a zstd frame which does not declare its content size."
+    chunker = zstandard.ZstdCompressor().chunker()
+    return b"".join(chunker.compress(data)) + b"".join(chunker.finish())
+
+
 def test_decode():
     """Test how responses are being decoded."""
     html_string = "<html><head/><body><div>ABC</div></body></html>"
@@ -306,13 +312,25 @@ def test_decode():
         compressed_strings.append(brotli.compress(html_string.encode("utf-8")))
     if HAS_ZSTD:
         compressed_strings.append(zstandard.compress(html_string.encode("utf-8")))
+        # servers compressing on the fly emit frames without a declared content
+        # size, and concatenated frames are equally valid
+        compressed_strings.append(zstd_stream_compress(html_string.encode("utf-8")))
+        compressed_strings.append(
+            zstandard.compress(html_string[:20].encode("utf-8")) + zstandard.compress(html_string[20:].encode("utf-8"))
+        )
 
     for compressed_string in compressed_strings:
         assert handle_compressed_file(compressed_string) == html_string.encode("utf-8")
         assert decode_file(compressed_string) == html_string
 
     # errors
-    for bad_file in ("äöüß", b"\x1f\x8b\x08abc", b"\x28\xb5\x2f\xfdabc"):
+    bad_files = ["äöüß", b"\x1f\x8b\x08abc", b"\x28\xb5\x2f\xfdabc"]
+    if HAS_ZSTD:
+        # a truncated frame decompresses to a prefix, which must not be mistaken
+        # for the whole document
+        truncated = zstd_stream_compress(html_string.encode("utf-8") * 500)
+        bad_files.append(truncated[: len(truncated) // 2])
+    for bad_file in bad_files:
         assert handle_compressed_file(bad_file) == bad_file
 
 
--- a/tests/downloads_tests.py
+++ b/tests/downloads_tests.py
@@ -330,6 +330,12 @@ def test_decode():
         # for the whole document
         truncated = zstd_stream_compress(html_string.encode("utf-8") * 500)
         bad_files.append(truncated[: len(truncated) // 2])
+        # reserved bit set in the frame header descriptor (the byte after the
+        # magic): rejected by the decompressor itself rather than by the
+        # end-of-frame check
+        frame = bytearray(zstandard.compress(html_string.encode("utf-8")))
+        frame[4] |= 0x08
+        bad_files.append(bytes(frame))
     for bad_file in bad_files:
         assert handle_compressed_file(bad_file) == bad_file
 
--- a/trafilatura/utils.py
+++ b/trafilatura/utils.py
@@ -143,6 +143,25 @@ RE_FILTER = re.compile(
 LINK_FARM_RATIO = 0.9
 
 
+def _decompress_zstd(filecontent: bytes) -> bytes | None:
+    """Decompress a Zstandard payload, or return None if it is not intact.
+    The one-shot API only accepts frames declaring their decompressed size in
+    the header, which servers compressing their responses on the fly omit."""
+    chunks: list[bytes] = []
+    remaining = filecontent
+    while remaining:
+        decompressor = zstandard.ZstdDecompressor().decompressobj()
+        try:
+            chunks.append(decompressor.decompress(remaining))
+        except zstandard.ZstdError:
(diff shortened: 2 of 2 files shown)


audi0417:
Pushed `2a8dd81` — rebased onto `2e1d38b`, plus the coverage gap and a title fix.

**Codecov patch coverage.** The two uncovered lines were the `except zstandard.ZstdError` branch in `_decompress_zstd`. None of the existing bad inputs reach it: a truncated frame and a zero-filled payload both decompress without raising and are caught by the `decompressor.eof` check instead, so the `except` was dead in the test suite even though it is not dead in practice.

The input that does raise is a frame with the reserved bit set in the frame header descriptor — the byte right after the magic, which RFC 8878 requires to be zero. That stays a valid trigger regardless of how the payload happens to be compressed, unlike corrupting a block header, which depends on the byte layout. Both failure modes are now covered separately, which is the point: they are different branches with the same outcome.

**Rebase.** The branch was 4 commits behind; it is now on `2e1d38b`. Rebased rather than merged, no conflicts.

**Title.** Dropped the `(#912)` — that was the issue number, but since this repo squash-merges, GitHub appends the PR number itself, so it would have landed as `... (#912) (#920)` and the first number would read as a PR that isn't one. `Fixes #912` in the description already does the linking. (Same cleanup you applied to #915 on the way in.)

**Verification on the rebased branch:** 354 passed / 1 skipped, `mypy -p trafilatura` clean, `ruff check` and `ruff format --check` clean, and `trafilatura/utils.py` lines 146–162 and 181–184 are fully covered locally.

adbar:
Thanks!

PR #922, opened 2026-08-28
fix(utils): use stream_reader for zstd decompression with unknown content size (#912)

### LINKED ISSUE #912: zstd responses without a declared content size are silently returned undecompressed

teddiesloco:
### Summary

Fixes #912.

When handling streamed responses compressed with Zstandard (`zstd`) without a declared frame content size (common for dynamic origins), one-shot `zstandard.decompress()` fails with `ZstdError: could not determine content size in frame header`. This error was caught and caused `handle_compressed_file` to return raw compressed bytes unchanged, resulting in failed extractions.

### Changes
- Replaced one-shot `zstandard.decompress(filecontent)` with `zstandard.ZstdDecompressor().stream_reader(filecontent).read()` in `trafilatura/utils.py`, enabling robust decompression of both fixed-size and streamed/unknown-size zstd frames.


adbar:
This PR is superseded by #920.

PR #924, opened 2026-08-28
fix: keep LaTeX math out of the CommonMark escaping pass

### LINKED ISSUE #923: Markdown output: the CommonMark escaping pass backslash-escapes LaTeX math, so $a_1 < a_2$ becomes $a\_1 \< a\_2$
## What happens

`xmltotxt()` rewrites LaTeX math first and escapes CommonMark metacharacters second:

```python
# trafilatura/xml.py:743-744 on 2e1d38b
_convert_math_tree(xmloutput)
_escape_inline_specials_tree(xmloutput)
```

The escaping pass added in #915 has no notion of math, so it runs over the `$...$` and
`$$...$$` spans the math pass has just produced and backslash-escapes every character in
`_INLINE_SPECIAL_CHARS` it finds there (`*`, `_`, `` ` ``, `~`, `[`, `]`, `<`). Inside a
formula those are LaTeX notation, not CommonMark: `_` is a subscript, `^*` is a conjugate,
`\left[ ... \right]` is a delimiter pair, `<` is a comparison. The output keeps the `$`
fences but the formula between them is no longer valid LaTeX, so it renders as literal
text rather than math.

Escaping the surrounding prose is right and I am not questioning that part. The formula
is the one region where a backslash is read by the math renderer instead of by CommonMark.

## Repro

`repro.py`:

```python

HTML = """<html><head><title>Relativity notes</title></head><body><article>
<p>For a sequence of measurements we write \\(a_1 &lt; a_2\\) whenever the second
detector records a larger deflection than the first one does, and the same ordering
holds for every later pair in the run that the laboratory published last spring.</p>
<p>The block form is also used in the literature, for example the identity below,
which appears in every textbook chapter that introduces the summation convention.</p>
<p>\\[ x_i = \\sum_j A_{ij} y_j \\]</p>
</article></body></html>"""

print(extract(HTML, output_format="markdown", favor_recall=True))
```

```
pip install -e .
python repro.py
```

## Expected

```
For a sequence of measurements we write $a_1 < a_2$ whenever ...

$$
x_i = \sum_j A_{ij} y_j
$$
```

## Actual

```
For a sequence of measurements we write $a\_1 \< a\_2$ whenever ...

$$
x\_i = \sum\_j A\_{ij} y\_j
$$
```

## Version and provenance

Reproduced in a clean `python:3.12-slim` container, `pip install -e .` from a fresh clone,
`trafilatura.__version__` 2.2.0, `trafilatura.xml.__file__` pointing at the checkout.

Bisected against the two commits around #915:

| commit | output |
| --- | --- |
| `e90b0ef` (parent of #915) | `$a_1 < a_2$`, block math intact |
| `2e1d38b` (current master) | `$a\_1 \< a\_2$`, block math escaped |

So this arrived with #915 rather than with the math support in #872.

## Why the suite did not catch it

The existing math assertions only use characters that are not in `_INLINE_SPECIAL_CHARS`:
`test_math_conversion` asserts on `$x^2$` and `$t$`, and `test_math_recovery`'s end to end
assertion is `$1.33\times 10^{-4}$`, where `^`, `{`, `}` and `-` are all left alone. Nothing
in the suite puts a subscript or a `<` inside a formula and then reads the Markdown, so the
escaping pass could be added without any test noticing.

## Suggested fix

The two passes cannot simply be swapped: escaping first turns a literal `[x]` into `\[x\]`,
which the math pass then reads as block math, so `[bracket]` in ordinary prose comes out as
a `$$` block. What works is one pass per text node that splits on the math delimiters,
escapes the prose regions and converts the math regions, so neither can see the other's
output.

I have that written and tested and will send it as a PR right after this.

ebarkhordar:
Fixes #923.

## Problem

`xmltotxt()` converted LaTeX math and then escaped CommonMark metacharacters, so the
escaping pass from #915 walked back over the `$...$` and `$$...$$` spans the math pass had
just produced and backslash-escaped the `*`, `_`, `` ` ``, `~`, `[`, `]` and `<` in them.
Inside a formula those are LaTeX, so `$a_1 < a_2$` came out as `$a\_1 \< a\_2$` and no
longer rendered as math.

## Change

Swapping the two passes does not work: escaping first turns a literal `[x]` in prose into
`\[x\]`, which the math pass then reads as block math, and `[bracket]` in an ordinary
sentence becomes a `$$` block. The two jobs have to agree on the regions before either of
them writes anything.

`_escape_and_convert_math()` splits each text node on the same delimiters `_convert_math()`
converts, escapes the prose parts and converts the math parts, so neither pass can see the
other's output. `_convert_math_tree()` and `_escape_inline_specials_tree()` collapse into
one walk with the same code-subtree skip rule they both had. `_escape_inline_specials()`
itself is unchanged, and it is still what escapes image alt text.

The unmatched-delimiter behaviour is unchanged: a lone `\(` is not a span, so its text is
escaped as ordinary prose.

## Tests

`test_math_is_not_commonmark_escaped` covers subscripts and `<` in inline math, a block
formula with braces, a star and `\left[ ... \right]`, prose escaping on both sides of a
formula, and the unmatched-delimiter case. The existing math assertions only used `^`, `{`,
`}` and `-`, none of which the escaping pass touches, which is why nothing failed when #915
landed.

## Verification

Clean `python:3.12-slim` container, `pip install -e ".[dev]"` from this branch.

- `pytest tests/unit_tests.py tests/realworld_tests.py tests/xml_tei_tests.py tests/filters_tests.py tests/baseline_tests.py tests/deduplication_tests.py tests/metadata_tests.py`: all pass.
- `mypy trafilatura/`: one pre-existing error in `utils.py:163`, unchanged by this branch and present on master.
- `ruff check .` and `ruff format --check .`: clean.
- `tests/cli_tests.py::test_sysoutput` fails in the container because it runs as root and can write `/root/forbidden/`. It fails the same way on unmodified master, so it is the container and not this change.


ebarkhordar:
One design question on this, since it is the part I would most want your read on. I merged `_convert_math_tree` and `_escape_inline_specials_tree` into a single walk because the two passes have to agree on where the math regions are before either one writes anything. The smaller alternative is to keep them as two walks and teach the escaping pass to skip `$...$` and `$$...$$`. I went the other way because that puts the delimiter set in two places that then have to stay in sync, but it is your call, and I would rather change the shape now than after you have read the whole diff.

adbar:
That works, thanks!

PR #926, opened 2026-09-06
Resolve sitemap locations by XML namespace

### LINKED ISSUE #440: xml namespace support in sitemaps
i ran into some sites that have a sitemap with xml prefix/namespace support
i now monkeypatched my code with following snippet to work around it

```
from trafilatura import sitemaps
sitemaps.LINK_REGEX = re.compile('<(?:\w+:)?loc>(?:<!\\[CDATA\\[)?(http.+?)(?:\\]\\]>)?</(?:\w+:)?loc>')                                                            
sitemaps.SITEMAP_FORMAT = re.compile(r'^.{0,5}<\?xml|<sitemap|<(?:\w+:)?urlset')
```

obviously this is not really maintanable on my side, so it would be better to get this upstreamed.
i can make a PR with this fix, but it's too simple in the sense that it does not limit and/or validate the namespaces.

eg when you would have a `<ns1:loc>` and also a `<ns2:loc>` in the same sitemap, this simple solution does not care and just grabs it all

it also only deals with `urlset` being namespaced atm

be-student:
Sitemaps with prefixed `urlset`, `sitemapindex` or `loc` tags currently lose their URLs. Resolve locations with the existing lxml dependency so prefixes follow their namespace bindings, including local rebindings, while extension locations such as `image:loc` are ignored. Keep legacy unnamespaced XML, CDATA, URL filtering and the location limit; disable entity resolution and network access in the parser.

Closes #440. Includes regression tests and a sitemap documentation update.

Validation: 21 sitemap tests pass (9 scenarios failed before the fix); repository-wide Ruff lint/format and mypy pass. Full pytest: 366 passed, 5 skipped, one CLI PATH failure; that test passed after activating the venv. The evaluation gate passes at both pinned floors. Local checks used Python 3.12 on macOS; CI's remaining platforms and optional dependency matrix were not run locally.


adbar:
@be-student The PR looks good and fixes issue 440. Please close the test coverage gaps.

Issue #585, opened 2024-05-04
Wrong links position in text from telegram post 


RedHotUnicorn:
Hello!

Found an peculiarity in extracting text from telegram posts.
So the link for example is [here](https://t.me/durov/272?embed=1&mode=tme) .

The text of the post is :
>Interestingly, most people [voted](https://t.me/durov/271) for Ukrainian to be the subject of my new post. So here you go — below is everything you wanted to know about the relationship of 
...

In html file the link just in div:
```html
<div>
Interestingly, most people <a href="https://t.me/durov/271" target="_blank" rel="noopener" onclick="return confirm('Open this link?\n\n'+this.href);">voted</a> for Ukrainian to be the subject of my new post. So here you go — below is everything you wanted to know about the relationship of
...
</div>
```

But after extracting markdown text contains two `\n` before each link:
>Interestingly, most people
>
>[voted](https://t.me/durov/271) for Ukrainian to be the subject of my new post. So here you go — below is everything you wanted to know about the relationship of

The code of extraction:
```python 
import trafilatura
url = "https://t.me/durov/272?embed=1&mode=tme"
html = trafilatura.fetch_url(url)

sent=trafilatura.extract(
    (
        html
    )
    , output_format='markdown'
    ,include_images=True
    # ,favor_precision=True
    ,include_formatting=True
    , include_links=True
)

print(sent)
```

I tried to use `favor_precision` option but its removing formatting and links.

RedHotUnicorn:
probably dup of https://github.com/adbar/trafilatura/issues/21

adbar:
Hi, I've never tried using Trafilatura on Telegram posts, I need to check what's going on.

ebarkhordar:
Still there on `2ba8f62`, with the t.me URL swapped for a fixture:

```python
from trafilatura import extract
p = "for Ukrainian to be the subject of my new post. " * 5
h = f'<html><body><div>Interestingly, most people <a href="https://t.me/durov/271">voted</a> {p}</div></body></html>'
extract(h, output_format="markdown", include_links=True, include_formatting=True)
# 'Interestingly, most people\n\n[voted](https://t.me/durov/271)\nfor Ukrainian to be ...'
```

Same break on 1.8.0, 1.9.0, 2.0.0 and 2.2.0. Wrapping that text in a `<p>` inside the same `<div>` keeps the sentence on one line.

PR #927, opened 2026-09-06
Fix paragraph breaks around inline links in text-bearing divs

emecii:
Addresses #585.

When Readability encounters a div containing ordinary text and an inline link,
its block-element pattern treats the anchor as a block. The subsequent pass
wraps the leading text and link tail in separate paragraphs, so Markdown gains
breaks around the link and XML leaves the ref outside its paragraph.

Allow paragraph conversion for these text-bearing divs while retaining block
detection inside links. Keep containers without loose text unchanged so their
Readability scoring is preserved. Explicitly retain the block/media elements
previously matched by the broad <a prefix.

Tests use original offline HTML and cover plain/nested-formatting anchors,
linked blocks, unchanged wrappers, and public Markdown/XML extraction with
fresh default size thresholds. Three new cases fail on the base; all eleven
pass with the fix.

Validation on macOS, Python 3.13.12, editable .[dev,all]:
- Offline pytest: 363 passed, one optional PyYAML test skipped; two live HTTP
  tests deselected. Network denied during execution.
- Repository-wide Ruff lint/format and mypy pass.
- Evaluation gate unchanged: fast 0.9195, fallback 0.9253.
- Of 990 fallback corpus outputs, one differs only in whitespace; annotated
  segment counts are unchanged.

This addresses the Readability conversion path; fast-mode link retention and
mixed-block layouts are outside this change. Other CI platforms were not run
locally.

Prepared with AI assistance and validated locally by the coding agent.
Human review remains pending.


adbar:
Thanks. Side note: this is indeed related to #585 but doesn't completely close it.

PR #928, opened 2026-09-07
fix: accept Extractor options in readerability check

### LINKED ISSUE #749: Review input type for `is_probably_readerable()` function
The options passed to this functions should be of the `Extractor` type and not `Any` or `dict`, the tests have to be rewritten accordingly:

```
def is_probably_readerable(html: HtmlElement, options: Optional[Extractor] = None) -> bool:
...
    if options:
        option_dict = {attr: getattr(options, attr, None) for attr in options.__slots__}
    else:
        option_dict = {}
```

emecii:
## Summary

- accept the project's `Extractor` object in `is_probably_readerable()`
- map only `min_extracted_size` to the heuristic's `min_content_length`
- retain the existing dictionary and `None` option behavior, with regression coverage and API documentation

Closes #749.

## Validation

- `python -m pytest` — 351 passed, 4 skipped
- `python -m pytest docs/docs_tests.py` — 2 passed
- `python tests/eval_gate.py` — fast F1 0.9195 and fallback F1 0.9253, both at their required floors
- `ruff check .` and `ruff format --check --diff .` — passed
- `mypy -p trafilatura` still reports the pre-existing `trafilatura/utils.py:163` `no-any-return` error; the identical error occurs with this change stashed


adbar:
`is_probably_readerable` is more a convenience function and not part of the main functionality.
This is a design decision, there are already a lot of different options and it can become confusing to add further ones. I'll park this PR for now.

PR #929, opened 2026-09-11
fix: drop Al Jazeera "Recommended Stories" widget from the text

cvl01:
## Summary
- Al Jazeera articles embed a `<section class="more-on">` widget ("Recommended Stories" heading + a linked list of other articles) between body paragraphs. No existing discard token matched `more-on`, so the heading and list items leaked into extracted output.
- Added `"more-on"` to `_RELATED_CLASS_TOKENS` in `trafilatura/xpaths.py` so the whole widget is pruned.

## Test plan
- [x] Added `tests/cache/aljazeera.com.malawians.html` and a regression case in `tests/realworld_tests.py` asserting the article text is kept while "Recommended Stories" and its list items are gone.
- [x] `pytest tests/realworld_tests.py tests/unit_tests.py` passes (194 tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adbar:
Thanks!

PR #930, opened 2026-09-13
fix: preserve empty table cells in XML output

### LINKED ISSUE #633: table extraction in xml
Hi,

Another inaccuracy issue in XML extraction for tables.

If the table contains one or more empty cells, the XML simply ignore it. For example, it makes a table with 3 row into 2 row.
`<table>
<row span="3">
<cell>a</cell>
<cell>b</cell>
</row>
<row span="3">
<cell>f</cell>
<cell>s</cell>
<cell>s</cell>
</row>
<row>
<cell>g</cell>
<cell>b</cell>
</row>
</table>`

It's better to extract empty cells as empty string or None to keep the layout correct.

`<table>
<row span="3">
<cell>a</cell>
<cell></cell>
<cell>b</cell>
</row>
<row span="3">
<cell>f</cell>
<cell>s</cell>
<cell>s</cell>
</row>
<row>
<cell>g</cell>
<cell>b</cell>
<cell>None</cell>
</row>
</table>`

Iams4kura:
XML and TEI output drops empty table cells even though the table extractor preserves them. For example, a row containing `a`, an empty cell, and `c` is serialized as only two cells, so `c` moves into the second column.

Preserve `cell` alongside `graphic` in both XML cleanup passes. Both changes are necessary: either cleanup would otherwise remove the placeholder. Text/Markdown extraction and table-content heuristics are unchanged.

Fixes #633. The earlier table handling work retained cells during extraction; this covers the remaining XML serialization path.

Regression coverage checks leading, middle, trailing, and colspan-placeholder cells through the public `extract()` API for both XML and TEI. The same eight cases fail on unchanged `master` at `eb04027c54cee100c4a6f203d4c31eff5b0e3ae6` and pass after the fix.

Validation on Python 3.12.13 with the pinned `dev,all` extras:

- `python -m pytest -q` — 374 passed, 1 skipped.
- `pytest tests/unit_tests.py tests/xml_tei_tests.py tests/realworld_tests.py -q` — 221 passed, 1 skipped.
- `mypy` — no issues in 22 source files.
- `ruff check .` and `ruff format --check .` — passed.
- `python tests/eval_gate.py` — fast F1 0.9195 and fallback F1 0.9253, both matching their pinned floors.


adbar:
@Iams4kura Could you define the set `{"graphic", "cell"}` as a constant in xml.py to avoid recasting it repeatedly?

Issue #774, opened 2025-01-07
Trafilatura fails to extract structured heading tags (h2, h3) 


LeMoussel:
Hi,

I encountered an issue where Trafilatura is unable to extract simple structured heading tags, such as `<h2>` or `<h3>`, from the website https://zonetuto.fr/.

Here is the Python code used:
```python
import urllib.request

class Trafilatura:
    def get_text(self, html: str) -> str:
        trafilatura_data = extract(
            filecontent=html,
            include_comments=False,
            favor_precision=True,
        )
        return trafilatura_data


if __name__ == "__main__":
    SITE_URL = "https://zonetuto.fr/"

    trafilatura = Trafilatura()
    txt_content = trafilatura.get_text(html_content)
    print(txt_content)
```
**Output**
```
Identifier et éliminer les fichiers en double peut vite devenir un casse-tête, surtout lorsque l’on accumule des documents, photos ou vidéos sur plusieurs supports. Si je vous écris cet article, c’est encore une fois après avoir eu besoin de mettre en pratique ce que je vais vous expliquer dans ce billet. J’ai une grosse machine … Lire la suite
Que le temps passe vite, cela faisait bien longtemps que je ne vous avais pas écrit un petit article sur l’intelligence artificielle. Ce n’est pas par manque d’envie, mais malheureusement encore une fois par manque de moments disponible pour vous écrire ici. Rassurez-vous je ne suis jamais très loin et j’essaie de garder le rythme … Lire la suite
L’univers du gaming en 2025 n’a jamais été aussi intense. Entre graphismes à couper le souffle, intelligence artificielle omniprésente et compétitions e-sportives qui attirent des millions de spectateurs, il est clair qu’avoir une tour gaming performante n’est plus une option, mais une nécessité. Alors, comment construire une bête de course sans exploser son portefeuille ? … Lire la suite
Le langage de programmation Python comporte beaucoup d’avantages. C’est facile à installer sur l’ensemble des systèmes d’exploitation, c’est assez rapide à apprendre et ça marche presque partout. Le Python est quasiment utilisable partout, que ce soit sur une machine Linux et ses différentes distributions, un Raspberry Pi, macOS ou encore sur Windows. Python est même … Lire la suite
[...]

PR #931, opened 2026-09-13
Keep every entry heading on listing pages

dualfroz:
On listing pages `(.//article)[1]` kept only the first entry, so the page escalated to justext and lost every heading.
The recall pass now tries the container of sibling articles first. This does not change calls with `favor_precision=True`, like the one in the report.

Refs #774


adbar:
@dualfroz Your PR is not necessarily beneficial in terms of recall. The matched container can be related-posts widget (e.g. "recommended stories") and not the main body.

PR #932, opened 2026-09-15
fix: extract JSON-LD metadata with array-valued types

xblwh:
`extract_metadata()` drops JSON-LD authors whose `@type` is an array, and skips article metadata when an enclosing type array starts with an unrecognized type, such as `["Thing", "NewsArticle"]` or `["CreativeWork", "Article"]`.

Normalize nested author types with the existing `as_list()` helper. For enclosing nodes, select the first recognized type across the existing article, page, publisher, and Person handling. Preserve input order among recognized types and the original first-element fallback when none is recognized.

The 47 added public-API cases cover scalar/array author types, enclosing article arrays, other supported node types, known-type precedence, and empty/unknown types.

Validation on macOS arm64 with the repository's pinned dependencies:

- Full pytest: **409 passed, 5 skipped** on Python 3.12 (`dev`); **413 passed, 1 skipped** on Python 3.14 (`dev,all`).
- Repository-wide mypy (22 source files), `ruff check .`, `ruff format --check .`, and `git diff --check` pass.
- Evaluation gate: fast F1 **0.9195**, fallback F1 **0.9253**, both matching the pinned floors.

Other OS/Python combinations were not run locally. Prepared with AI assistance from Codex.


trafilatura/json_metadata.py
```diff
@@ -158,7 +158,7 @@ def process_parent(parent: Any, metadata: Document) -> Document:
                 for author in as_list(list_authors):
                     if isinstance(author, str):
                         author = {"name": author}
-                    if "@type" not in author or author["@type"] == "Person":
+                    if "@type" not in author or "Person" in as_list(author["@type"]):
```
ebarkhordar, line 169: I ran this against 3e72f888 in a clean container (python:3.12-slim, the package installed from this head).

The fix is right and `as_list` is the correct tool for it. It cannot run, though, when the enclosing article node carries an array `@type` whose article type is not first, because line 130 still selects only the first element:

```python
content_type = content["@type"][0] if isinstance(content["@type"], list) else content["@type"]
```

Measured at this head, with the author sub-node a plain `{"@type": "Person", "name": "Jane Doe"}` dict in every case:

```
@type='NewsArticle'              author='Jane Doe'  pagetype='newsarticle'
@type=['NewsArticle']            author='Jane Doe'  pagetype='newsarticle'
@type=['NewsArticle','Thing']    author='Jane Doe'  pagetype='newsarticle'
@type=['Thing','NewsArticle']    author=None        pagetype=None
@type=['CreativeWork','Article'] author=None        pagetype=None
```

The last two never reach the `content_type in JSON_ARTICLE_SCHEMA` branch, so the node is skipped whole and pagetype and sitename go with the author. The comment above line 130 reads "some websites are using ['Person'] as type", which covers a one-element array but not a longer one. Your own new test covers `["Thing", "Person"]` for an author, and that is the same ordering this misses one level up.

One edit in the same style: choose the first element of `as_list(content["@type"])` that falls in one of the known schema sets, and keep the present first-element behaviour when none does.

I did not run the F1 evaluation gate.

xblwh: Thanks for the detailed reproduction. Fixed in 419ed6c: `process_parent()` now selects the first recognized type, including Person, while preserving the order among recognized types and falling back to the first element when none matches.

Added 29 public-API cases covering both examples, scalar/array author types, other node types, and existing type precedence. They give 10 failures and 19 passes on the previous head, then all pass with this update. Full pytest passes on Python 3.12 (409 passed, 5 skipped) and 3.14 (413 passed, 1 skipped); mypy and Ruff pass. I also ran the F1 gate: fast 0.9195 and fallback 0.9253, matching both floors.

I checked sitename too: with a dictionary-valued publisher it was already retained because publisher extraction runs before type selection. The regression tests verify that it stays intact.

ebarkhordar: Confirmed at 419ed6c. The two ordering cases from my comment now resolve, and I compared the whole `@type` matrix against 3e72f888: every difference is a node that previously extracted nothing, so the new selection only fires where the first-element version was inert. Nothing that already resolved changed.

Your `isinstance(schema_type, str)` filter also picks up a case for free. `"@type": [null, "NewsArticle"]` and `[123, "NewsArticle"]` raised `AttributeError` on the old line and now extract normally.

I did not re-run the F1 gate.

adbar:
Thanks!

PR #937, opened 2026-09-22
fix: do not duplicate the first item of a nested list

SulimanAbdulrazzaq:
When the first item of a nested list has child elements, `handle_lists()` emits it a second time as an extra item of the parent list. The stray item is part of the extracted tree, so it shows up in the txt, Markdown and XML output alike.

Example: `<ul><li>a<ul><li>b<ul><li>c</li></ul></li><li>b2</li></ul></li><li>d</li></ul>`. The strings below are `extract()` txt output from a minimal reproduction with the minimum sizes set to 0; the leading intro paragraph of the test page is left out:

```
master:     '- a\n  - b\n    - c\n  - b2\n- b\n- d'
this PR:    '- a\n  - b\n    - c\n  - b2\n- d'
```

On master the XML output also contains a second, top-level `<item>b</item>`. With `include_links=True`, a first nested item like `<li>see <a href="https://example.org/x">link</a></li>` leaves a partial copy behind: master gives `'- a\n  - see [link](https://example.org/x)\n  - c\n- see \n- d'`.

**Cause.** `handle_lists()` walks every descendant `<item>` of a list. It relies on the recursive call for a nested list to rename that list's items to `"done"`, so that the outer loop skips them. The problem is that lxml's iterator fetches the next matching element before the loop body runs. The first item of the nested list is therefore still yielded after the recursive call has processed and renamed it. For a leaf item this is harmless, because `process_node()` returns `None` for `"done"` elements and the item is dropped. An item with child elements goes through `process_nested_elements()` instead, which has no such check, so its text is appended to the outer list again.

**Fix.** Skip items that are already `"done"` at the top of the loop. The table cell loop in the same module already does this.

**Tests.** `test_nested_list_first_item_not_duplicated` covers three cases:
- `handle_lists()` called directly on a nested list;
- the txt and XML output for a three-level list;
- the txt output for a nested first item that contains a link.

It fails on master (`assert ['a', 'b'] == ['a']`) and passes with the fix.

Checks, run on Windows 11 with Python 3.11 and `-e ".[dev]"`:
- `python -m pytest`: 429 passed, 5 skipped (the skipped tests need pycurl, PyYAML or py3langid, which were not installed)
- `mypy -p trafilatura`: no issues found in 22 source files
- `ruff check .` and `ruff format --check --diff .`: pass
- `python tests/eval_gate.py`: fast F1 0.9195 and fallback F1 0.9253, identical on master and with this change

Not run locally: `docs/docs_tests.py` (it needs the docs extra, and this PR does not touch the docs), and the other CI platforms.


SulimanAbdulrazzaq:
@adbar when you have a moment, could you take a look at this one? When the first item of a nested list has child elements, `handle_lists()` emits it a second time as an extra item of the parent list, so it appears twice in the txt, Markdown and XML output. The branch merges cleanly and CI is green.

